PR #11, opened 2025-06-09
#4 OIP FastAPI - Model Metadata and Model Ready endpoints

### LINKED ISSUE #4: Feature: Create the OIP Model Metadata request for the FastAPI implementation
## Description
In issue #1, the initial FastAPI implementation of the Open Inference Protocol was created. That ticket handled generating the project and the infer endpoint. This ticket should pick up from there and implement the model metadata and model ready endpoints. This includes the endpoints, data types, and the handler.

## DOD
- [x] Update habushu to version 3
  - [x] If the upgrade does not automatically set the project to use Ruff then do so
- [x] Create an interface for each of the model [REST V2 endpoints](https://kserve.github.io/website/latest/modelserving/data_plane/v2_protocol/#httprest) and their dataplane objects
  - [x] `v2/models/<model_name>[/versions/<model_version>]`
  - [x] `v2/models/<model_name>[/versions/]/ready`
  - ---Split to another ticket---
  - [ ] `v2/health/ready`
  - [ ] `v2/health/live`
  - [ ] `v2`
- [x] Update docs in aissemble-open-inference-api-fastapi README if needed to outlining the intended use of these endpoints. We dont want to duplicate the OIP docs however

## Test Strategy/Script
1. Build the aissemble-open-inference-protocol project - `mvnd clean install`
2. In a new test directory, create a temp venv and pip install aissemble-open-inference-protocol-fastapi. Note because the package is not pushed, we have to pip install the whl:

```
mkdir test-oip && cd test-oip
python -m venv venv
source venv/bin/activate
pip install {path/to}/aissemble-open-inference-protocol/aissemble-open-inference-protocol-fastapi/dist/aissemble_open_inference_protocol_fastapi-1.0.0.dev0-py3-none-any.whl --force
```

4. Create a `main.py` file with a simple request handler
```
from typing import Optional

from aissemble_open_inference_protocol_fastapi.aissemble_oip_fastapi import (
    AissembleOIPFastAPI,
)
from aissemble_open_inference_protocol_fastapi.handlers.dataplane import (
    DataplaneHandler,
)
from aissemble_open_inference_protocol_fastapi.types.dataplane import (
    InferenceRequest,
    InferenceResponse,
    ModelMetadataResponse,
    MetadataTensor,
    ModelReadyResponse,
)


class Handler(DataplaneHandler):
    def __init__(self):
        super().__init__()

    def infer(
        self,
        payload: InferenceRequest,
        model_name: str,
        model_version: Optional[str] = None,
    ) -> InferenceResponse:
        return InferenceResponse(
            model_name=model_name, model_version=model_version, id="id", outputs=[]
        )

    def model_metadata(
        self,
        model_name: str,
        model_version: Optional[str] = None,
    ) -> ModelMetadataResponse:
        # Return a stub ModelMetadataResponse
        return ModelMetadataResponse(
            name=model_name,
            versions=[model_version] if model_version else None,
            platform="python",
            inputs=[MetadataTensor(name="input", datatype="FP32", shape=[1])],
            outputs=[
                MetadataTensor(name="output", datatype="FP32", shape=[1])
            ],
        )

    def model_ready(
        self,
        model_name: str,
        model_version: Optional[str] = None,
    ) -> ModelReadyResponse:
        # Testing: always ready
        return ModelReadyResponse(name=model_name, ready=True)


app = AissembleOIPFastAPI(Handler).app
```
5. Start the server - `fastapi dev main.py`
6. View the API docs - http://127.0.0.1:8000/docs
7. In the default section, ensure you see the endpoints:
    - `v2/models/{model_name}`
    - `v2/models/{model_name}/versions/{model_version}`
    - `v2/models/{model_name}/ready`
    - `v2/models/{model_name}/versions/{model_version}/ready`
8. In the schemas section, ensure the schemas for each datatype match the [OIP Spec](https://github.com/kserve/open-inference-protocol/blob/main/specification/protocol/inference_rest.md#model-metadata-response-json-object):
    - `ModelMetadataResponse` (in docs, `metdata_tensor` will show up as `array<object>`. Expand "items" and ensure fields match `metadata_tensor` format)
        - ```
           $metadata_model_response =
            {
              "name" : $string,
              "versions" : [ $string, ... ] #optional,
              "platform" : $string,
              "inputs" : [ $metadata_tensor, ... ],
              "outputs" : [ $metadata_tensor, ... ]
            }
            ``` 
    - `MetadataTensor`
        - ````
           $metadata_tensor =
            {
              "name" : $string,
              "datatype" : $string, # Datatype enum
              "shape" : [ $number, ... ]
            } 
          ````
    - `MetadataModelErrorResponse`
        - ```
           $metadata_model_error_response =
            {
              "error": $string
            }
           ```  
    - `ModelReadyResponse`
        - ```
           $ready_model_response =
            {
              "name" : $string,
              "ready": $bool
            }
           ```  
9. Use the "Try it out" to send requests to all 4 endpoints
    - In the default section, expand one of the endpoints
    - Select "Try it out"
    - Fill in model and model version if applicable
    - Select execute
        - Server response should show Code 200 and model name should be in the output with other possible responses listed below
    - Repeat this process for all endpoints
10. Update the app constructor to use the default aiSSEMBLE OIP FastAPI request handler
    - In the `main.py` file, remove the custom handler so the app line looks like: `app = AissembleOIPFastAPI().app`
11. Restart the server - `fastapi dev main.py`
12. Repeat step 9
    - It should now return with a 501 not implemented

Clean up
13. Deactivate the venv: `deactivate`
14. Remove the test directory: cd up a level and run `rm -r test-oip` 

[Open Inference Protocol](https://github.com/kserve/open-inference-protocol?tab=readme-ov-file#open-inference-protocol-specification)

colinpalmer-pro:
Changes include

- Updated habushu to v3.0.0
- Model metadata enpoint
- Model ready endpoint
- Initial tests to ensure endpoints are reachable in default implementation (501) and that endpoints are present in API documentation

### DIFF
--- a/aissemble-open-inference-protocol-fastapi/README.md
+++ b/aissemble-open-inference-protocol-fastapi/README.md
@@ -23,15 +23,25 @@ fastapi dev main.py
 
 View the routes by going to http://127.0.0.1:8000/docs.
 
-## Implementing Inference Handler
-The endpoints will call a [default handler](src/aissemble_open_inference_protocol_fastapi/handlers/default_handler.py) that will return 501 not implemented. To make a handler, create your class and extend the abstract base class [dataplane.py](src/aissemble_open_inference_protocol_fastapi/handlers/dataplane.py). Then just pass your class into our AissembleOIPFastAPI constructor.
+## Implementing a Handler
+The endpoints will call a [default handler](src/aissemble_open_inference_protocol_fastapi/handlers/default_handler.py) that will return 501 not implemented. To make a handler, create your class and extend the abstract base class [dataplane.py](src/aissemble_open_inference_protocol_fastapi/handlers/dataplane.py). Then pass your class into the AissembleOIPFastAPI constructor.
 
 ### Example of Usage with A Handler
 Create your custom handler class with:
 ```python
 from typing import Optional
-from aissemble_open_inference_protocol_fastapi.handlers.dataplane import DataplaneHandler
-from aissemble_open_inference_protocol_fastapi.types.dataplane import InferenceRequest, InferenceResponse
+
+from aissemble_open_inference_protocol_fastapi.handlers.dataplane import (
+    DataplaneHandler,
+)
+from aissemble_open_inference_protocol_fastapi.types.dataplane import (
+    InferenceRequest,
+    InferenceResponse,
+    ModelMetadataResponse,
+    MetadataTensor,
+    ModelReadyResponse,
+)
+
 
 class MyHandler(DataplaneHandler):
     def __init__(self):
@@ -43,7 +53,33 @@ class MyHandler(DataplaneHandler):
             model_name: str,
             model_version: Optional[str] = None,
     ) -> InferenceResponse:
-        return InferenceResponse(model_name=model_name, model_version=model_version, outputs=[])
+        return InferenceResponse(
+            model_name=model_name, model_version=model_version, id="id", outputs=[]
+        )
+
+    def model_metadata(
+            self,
+            model_name: str,
+            model_version: Optional[str] = None,
+    ) -> ModelMetadataResponse:
+        # Return a stub ModelMetadataResponse
+        return ModelMetadataResponse(
+            name=model_name,
+            versions=[model_version] if model_version else None,
+            platform="python",
+            inputs=[MetadataTensor(name="input", datatype="FP32", shape=[1])],
+            outputs=[
+                MetadataTensor(name="output", datatype="FP32", shape=[1])
+            ],
+        )
+
+    def model_ready(
+            self,
+            model_name: str,
+            model_version: Optional[str] = None,
+    ) -> ModelReadyResponse:
+        # Testing: always ready
+        return ModelReadyResponse(name=model_name, ready=True)
 ```
 
 Use aissemble-open-inference-protocol-fastapi to create a FastAPI app and pass it `MyHandler`
--- a/aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
+++ b/aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
@@ -15,6 +15,9 @@ from aissemble_open_inference_protocol_fastapi.handlers.default_handler import (
 from aissemble_open_inference_protocol_fastapi.types.dataplane import (
     InferenceRequest,
     InferenceResponse,
+    ModelMetadataResponse,
+    ModelMetadataErrorResponse,
+    ModelReadyResponse,
 )
 
 router = APIRouter(
@@ -35,9 +38,8 @@ def infer_model(
     payload: InferenceRequest,
     handler: DefaultHandler = Depends(DefaultHandler),
 ) -> InferenceResponse:
-    """## performs inference on a model.
-
-    Returns: the prediction results
+    """
+    Perform inference using the specified model and return the prediction results.
     """
     return handler.infer(model_name=model_name, payload=payload)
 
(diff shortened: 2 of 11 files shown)


aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -62,3 +65,123 @@ async def infer_model_version(
...
+            "model": ModelMetadataErrorResponse,
+            "description": "Returned if the model or model version is not found.",
+        },
+    },
+)
+def model_metadata(
+    model_name: str,
+    handler: DefaultHandler = Depends(DefaultHandler),
+) -> ModelMetadataResponse:
+    """
```
colinpalmer-pro, line 89: i: Trying a different docstring format here (should align all of them eventually). Swagger/OpenAPI supports markdown so in the UI docs, this ends up looking like:
image

Can view difference to the simpler infer endpoints. Fine with either way, just providing a different option to think about

aissemble-open-inference-protocol-fastapi/tests/features/fastapi.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Test FastAPI OIP Implementation
```
colinpalmer-pro, line 1: i: First scenario confirms runtime behavior (endpoints exist in default handler). Second scenario ensures the API shows the correct endpoints. Open to suggestions for other options

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
cwoods-cpointe: s: We should update inference to also have this format so they stay aligned

cwoods-cpointe: C: I dont think we want to have the handler show up in the api docs because its not rest related

colinpalmer-pro: Ah thats a good catch. In that case I think we should go with the simpler one liner we have with the infer endpoints. Just removing the handler argument makes sense for the public API but then makes it inconsistent with the internal implementation.

- [x] Update endpoint docstrings to remove handler reference and provide single line function summary

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/types/dataplane.py
```diff
@@ -82,3 +82,26 @@ class InferenceRequest(BaseModel):
     parameters: Optional[Parameters] = None
     inputs: List[RequestInput]
     outputs: Optional[List[RequestOutput]] = None
+
+
+class MetadataTensor(BaseModel):
+    name: str
+    datatype: str
```
cwoods-cpointe: c: we should use the same inference datatype

aissemble-open-inference-protocol-fastapi/README.md
```diff
@@ -23,15 +23,25 @@ fastapi dev main.py
 
 View the routes by going to http://127.0.0.1:8000/docs.
 
-## Implementing Inference Handler
-The endpoints will call a [default handler](src/aissemble_open_inference_protocol_fastapi/handlers/default_handler.py) that will return 501 not implemented. To make a handler, create your class and extend the abstract base class [dataplane.py](src/aissemble_open_inference_protocol_fastapi/handlers/dataplane.py). Then just pass your class into our AissembleOIPFastAPI constructor.
+## Implementing a Custom Handler
```
epwilkins: A: This makes it sound optional
```suggestion
## Implementing a Handler
```

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/types/dataplane.py
colinpalmer-pro: - [x] Update `datatype` to be type `Datatype`

aaron-gary (approved):

cwoods-cpointe (approved):

Issue #15, opened 2025-06-11
Feature: Add example for FastAPI REST interface


chang-annie:
## Description
Having examples baked into the project would help users coming to the project understand how `aissemble-open-inference-protocol` is meant to work with their existing machine learning models/code.  We should create an `examples` directory containing samples that follow a simple process like: pull in a model, have backend to interface with model, hook into `aissemble-open-inference-protocol` and run OIP-compliant inferencing. This ticket should create the initial examples directory & add the FastAPI REST interface example

## DOD
- [x] Create Example folder to create Fast API Rest interface example
- [x] Create Model, inferencing service and hook into OIP to demonstrate how it can use OIP to kick off inferencing endpoint. 
- [x] Support inference endpoint in the example
      -------split--------
  - [ ] Support Rest of endpoint 
  - [ ] Some sort of CICD (github action or build action (need discovery) )  implementation so that it can make sure it doesn't fail on each iteration of PR. 
- [x] Add README within example explaining usage
- [x] Update README from root to add Example section to redirect README of examples

## Test Strategy/Script
1. cd into  aissemble-open-inference-protocol project 
2. Build the aissemble-open-inference-protocol project - `mvnd clean install`
3. Start the server - 
`cd ./examples/fast-api-inference`
`poetry run fastapi dev ./src/fast-api-inference/main.py`
NOTE: wait until log says   `Application startup complete.
`
4. On another fresh terminal, run Following cURL command to send inference request
```
curl -H 'Content-Type: application/json' \
      -d '{"id" : "1",
  "inputs" : [{
      "name" : "sample",
      "shape" : [1],
      "datatype"  : "FP32",
      "data" : [ [ 25], [ 12 ] ]
  }]}' \
      -X POST \
      http://127.0.0.1:8000/v2/models/convert_celsius_to_fahrenheit/infer
```

5. Verify POST Request succeeded and following response shows up.
```
{
  "model_name": "convert_celsius_to_fahrenheit",
  "model_version": null,
  "id": "1",
  "parameters": null,
  "outputs": [
    {
      "name": "convert_celsius_to_fahrenheit",
      "shape": [
        1
      ],
      "datatype": "FP32",
      "parameters": null,
      "data": [
        [
          [
            77.19401550292969
          ],
          [
            53.622467041015625
          ]
        ]
      ]
    }
  ]
}
```

jaebchoi:
DOD Completed with @cwoods-cpointe

[...]

PR #21, opened 2025-06-20
[#15] create example fastAPI REST Interface

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/15


aissemble-open-inference-protocol-fastapi/example/model/model.py
```diff
@@ -0,0 +1,20 @@
...
+
+input_array = np.array(input_a, dtype=float)
+output_array = np.array(output_a, dtype=float)
+
+model = tf.keras.Sequential()
+model.add(tf.keras.layers.Input(shape=[1]))
+model.add(tf.keras.layers.Dense(units=1))
+model.compile(optimizer=tf.keras.optimizers.SGD(), loss=tf.keras.losses.MeanSquaredError())
+model.layers[0].get_weights()
+model.fit(input_array, output_array, epochs=1000)
```
jaebchoi: I: .keras file is just saved model of this model.py so that I can load model from keras file.

aissemble-open-inference-protocol-fastapi/example/README.md
epwilkins, line 1: C: I think we want this to be an `examples` directory, that then has something like `aissemble-oip-fast-api-example` like [Habushu](https://github.com/TechnologyBrewery/habushu/tree/dev/examples) has. I'm not sure if we will have multiple FastAPI examples or not and that might change the structure/naming/

aissemble-open-inference-protocol-fastapi/example/main.py
```diff
@@ -0,0 +1,90 @@
...
+    MetadataTensor,
+    ModelReadyResponse,
+    ServerReadyResponse,
+    ServerLiveResponse,
+    ServerMetadataResponse,
+    ResponseOutput,
+    Datatype,
+)
+
+class Handler(DataplaneHandler):
```
epwilkins: S: We should add docstrings to help explain the example

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Setting up the Project
+### Create Temporary virtual Environment
+we need to stand up temporary virtual environment and install aissemble-open-inference-protocol first
+NOTE: because the package is not pushed to testpypi, we have to pip install the whl:
+
+```sh
+python -m venv venv
+source venv/bin/activate
+pip install {path/to}/aissemble-open-inference-protocol/aissemble-open-inference-protocol-fastapi/dist/aissemble_open_inference_protocol_fastapi-1.0.0.dev0-py3-none-any.whl --force
```
woods-cpointe: c: we want this to be the actual install and not the path to the whl

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
+# aiSSEMBLE Open Inference Protocol FastAPI Example
+This example demonstrates how to hook model into Open Interface Protocol and use FastAPI to implement inference endpoint.
+
+## Setting up the Project
+### Create Temporary virtual Environment
```
woods-cpointe:
```suggestion
### Create Virtual Environment
```

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
...
+```sh
+python -m venv venv
+source venv/bin/activate
+pip install {path/to}/aissemble-open-inference-protocol/aissemble-open-inference-protocol-fastapi/dist/aissemble_open_inference_protocol_fastapi-1.0.0.dev0-py3-none-any.whl --force
+```
+
+### Install Necessary Library
+We also need to install numpy and tensorflow for the ML model.
+```sh
+pip install numpy
```
epwilkins: S: This should be a proper project with a pom and pyproject.toml that handles dependencies and project building.  It might make sense for us to also create a common goal or profile for starting the example servers so it's always the same command.

aissemble-open-inference-protocol-fastapi/example/README.md
woods-cpointe: The package is in testpypi but regardless we want to present the actual pip install

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
...
+pip install tensorflow
+```
+
+## Running Example FastAPI
+
+### Run FASTAPI development Server
```
epwilkins: A: I don't think we need separate sections. Less is more with these examples. We want the document to be as small as is possible to demonstrate the feature that we're making an example for.  This could easily be a 2 step list of instructions under _Running the Example_

aissemble-open-inference-protocol-fastapi/example/README.md
cwoods-cpointe: +1. The examples project should be a standard habushu project like the others in repo

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
...
+fastapi dev main.py
+```
+
+### Run Inference Endpoint
+
+Run following Endpoint with your choice of API Testing software
+
+NOTE: model name used for this example is called sample_convert
+```sh
+POST http://127.0.0.1:8000/v2/models/{model_name}/infer
```
epwilkins: S: We should give them a command that they can run so they don't have to do the work of choosing a REST client and mapping this info into a client command.

cwoods-cpointe: s: we know the model_name so we should just put it in this post and remove the NOTE

aissemble-open-inference-protocol-fastapi/example/main.py
```diff
@@ -0,0 +1,90 @@
...
+        super().__init__()
+
+    def infer(
+            self,
+            payload: InferenceRequest,
+            model_name: str,
+            model_version: Optional[str] = None,
+    ) -> InferenceResponse:
+        model = load_model("model/" + model_name + ".keras")
+        temp_f = model.predict(np.array(payload.inputs[0].data))
```
epwilkins: S: Because this is an example, we need to use very explicit variable names that makes the code super easy to read and follow.

aissemble-open-inference-protocol-fastapi/example/README.md
```diff
@@ -0,0 +1,50 @@
+# aiSSEMBLE Open Inference Protocol FastAPI Example
+This example demonstrates how to hook model into Open Interface Protocol and use FastAPI to implement inference endpoint.
```
epwilkins: A: It would be helpful to explain the sample model we're using and what the inference endpoint is actually producing.

aissemble-open-inference-protocol-fastapi/example/main.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def model_ready(
+            self,
+            model_name: str,
+            model_version: Optional[str] = None,
+    ) -> ModelReadyResponse:
+        # Testing: always ready
+        return ModelReadyResponse(name=model_name, ready=True)
+
+    def server_ready(
```
epwilkins: Q: Do we have to implement these? If so, it might make sense (in a follow-on) to make some base handler that returns Ready/metadata responses where we can. Obviously some things like the infer endpoint or the model metadata just _require_ project info and so returning a 5xx makes sense. But for `server_ready` I'd think most of the time if the FastAPI service is routing traffic to this function, then the server is ready.

aissemble-open-inference-protocol-fastapi/example/README.md
jaebchoi: sounds good, Ill update to examples directory!

aissemble-open-inference-protocol-fastapi/example/model/model.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+import tensorflow as tf
+
+input_a = []
+output_a = []
+
+for c in range(0, 16):
```
epwilkins: Q/S: Is this creating the model? If so, I'm not sure it should be in the example.  I'd think you wouldn't be creating a model in the inference endpoint, and just be serving an existing model. As such, this just adds noise to the example that makes it harder to follow. Examples should be very minimal and focus in on just demonstrating the feature.  I think it's fine to say we have some example model that already exists and it's assumed the reader will know how to get the model they want to serve.

aissemble-open-inference-protocol-fastapi/example/model/sample_convert.keras
epwilkins, line 1: A: This file could have a more meaningful name.

epwilkins (approved):

aissemble-open-inference-protocol-fastapi/example/README.md
jaebchoi: Oh I see realized that testpypi work is done, Ill update that!

aissemble-open-inference-protocol-fastapi/example/main.py
cwoods-cpointe: Yeah I think that makes sense for server_ready and server_live. Maybe for server metadata as well

aissemble-open-inference-protocol-fastapi/example/README.md
jaebchoi: Sounds good, Ill remove ### Run FASTAPI development Server

aissemble-open-inference-protocol-fastapi/example/main.py
```diff
@@ -0,0 +1,90 @@
...
+            model_name=model_name, model_version=model_version, id=payload.id, outputs=[ResponseOutput(name=model_name, shape=payload.inputs[0].shape, datatype=Datatype.FP32, data=[temp_f_list] )]
+        )
+
+    def model_metadata(
+            self,
+            model_name: str,
+            model_version: Optional[str] = None,
+    ) -> ModelMetadataResponse:
+        # Return a stub ModelMetadataResponse
+        return ModelMetadataResponse(
```
cwoods-cpointe: Q/s: does the tf.keras model object have model metadata data we can use?

aissemble-open-inference-protocol-fastapi/example/README.md
jaebchoi: I see, I can update this to curl command instead

jaebchoi: Sounds good, Ill update to add actual model name

aissemble-open-inference-protocol-fastapi/example/model/model.py
jaebchoi: Yup, this is actual code for creating a model, and saved this model as .keras file and main.py will basically load models from keras file. So actual example would never touch model file, This is intended to save later in case we ever modify models, but now that I am thinking about this might cause some confusion. So Ill delete this model.py and keep it for myself and just have .keras file pushed.

aissemble-open-inference-protocol-fastapi/example/main.py
jaebchoi: keras sequential model( that I am using ) has inputs and outputs https://www.tensorflow.org/api_docs/python/tf/keras/Model#with_the_sequential_class

That can retrieve datatype and shape but no platform.. given that keras is only supporting python maybe we can assume?

aissemble-open-inference-protocol-fastapi/example/main.py
jaebchoi: Makes sense I had to implement all of endpoint because I inherited Handler from [DataplaneHandler](https://github.com/boozallen/aissemble-open-inference-protocol/blob/dev/aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/handlers/dataplane.py#L24-L60) which is all abstract class which makes it required to implement all. I can inherit from [DefaultHandler](https://github.com/boozallen/aissemble-open-inference-protocol/blob/4e8e30b98a73f880586320326f9f4957d256bbd0/aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/handlers/default_handler.py#L27-L75) so that umimplemented endpoint always go default handler.

cwoods-cpointe: We should do what our lib expects users to do which is extend the abc. Good find in this pain point. We can update the abc in a follow on

aissemble-open-inference-protocol-fastapi/example/README.md
jaebchoi: Sounds good, updated pom and pyproject and added profile

aissemble-open-inference-protocol-fastapi/examples/fast-api-inference/main.py
```diff
@@ -0,0 +1,96 @@
+import numpy as np
```
cwoods-cpointe: c: This should be under src/fast-api-inference/

jaebchoi: updated!

examples/fast-api-inference/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+]
+license = { text = "Booz Allen Public License v1.0" }
+license-files = ["LICENSE*"]
+dynamic = ["version", "dependencies"]
+
+[tool.poetry.dependencies]
+fastapi = {extras = ["standard"], version = "^0.115.12"}
+tensorflow = "^2.19.0"
+krausening = "^21"
+aissemble-open-inference-protocol-fastapi = "^1.0.0.dev1750572548"
```
cwoods-cpointe: C: This would have the example running against the latest pushed artifact and not the current code base. This means issues with the examples wouldnt be caught until after the build and push. We want to point the dep to the actual module

jaebchoi: I see, I can do local relative path of aissemble-open-inference-protocol-fastapi

examples/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        <groupId>com.boozallen.aissemble</groupId>
+        <artifactId>aissemble-open-inference-protocol</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>examples</artifactId>
+    <packaging>pom</packaging>
+
+    <name>aiSSEMBLE::Open Inference Protocol Examples</name>
+
```
cwoods-cpointe, line 21: c: needs to have module fast-api-inference and the root needs to include the examples module

examples/fast-api-inference/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+    "Framework :: FastAPI",
+    "Intended Audience :: Developers"
+]
+license = { text = "Booz Allen Public License v1.0" }
+license-files = ["LICENSE*"]
+dynamic = ["version", "dependencies"]
+
+[tool.poetry.dependencies]
+fastapi = {extras = ["standard"], version = "^0.115.12"}
+tensorflow = "^2.19.0"
```
cwoods-cpointe: tensorflow 2.16.x is the latest version that offers mac x86 support. Please use that version if possible

examples/pom.xml
jaebchoi: updated modules as listed!

aissemble-open-inference-protocol-fastapi/README.md
```diff
@@ -123,4 +123,7 @@ The `content_type` feature lets you declare, at either request time or in model
 
 **Declaring Content Types:**
 - **Input Level**: Include a `content_type` field under each `inputs[].parameters` in the request JSON. This will apply the content type to the particular input only.
-- **Model Metadata Level**: This will apply the content type to the entire set of inputs.
\ No newline at end of file
+- **Model Metadata Level**: This will apply the content type to the entire set of inputs.
+
+## Example Usage with Actual ML Model ##
+Please see the [example](./example/README.md) folder dedicated for detailed information on how to integrate models with Fast API OIP.
```
cwoods-cpointe: a: This link needs to be updated to the blob so the links in the pypi page are valid

examples/fast-api-inference/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+    "Framework :: FastAPI",
+    "Intended Audience :: Developers"
+]
+license = { text = "Booz Allen Public License v1.0" }
+license-files = ["LICENSE*"]
+dynamic = ["version", "dependencies"]
+
+[tool.poetry.dependencies]
+fastapi = {extras = ["standard"], version = "^0.115.12"}
+tensorflow = "^2.16.0"
```
cwoods-cpointe: a: The `^` still brings 2.19.0. 
```suggestion
tensorflow = ">=2.16.0,<2.17.0"
```

aissemble-open-inference-protocol-fastapi/README.md
jaebchoi: good catch, updated!

cwoods-cpointe (approved):

Issue #12, opened 2025-06-10
Feature: Add implementation to type the request and response payload


chang-annie:
## Description
This is a follow-on to #3. 
During the initial implementation of the FastAPI REST interface it was noted that the OIP implementation for input and output is too general. To make it easier to work with it would be good to add data typing. https://github.com/boozallen/aissemble-open-inference-protocol/pull/2#discussion_r2098052098 is the PR comment that inspired this issue.
We added a section to the `aissemble-open-inference-protocol-fastapi` readme about content types but have not implemented it yet. Within this ticket, we should detail the level of typing we want to provide and begin implementation

## DOD
- [x] remove headers object - no demand signal and not being used currently
- [x] implement content type decoding/encoding (`bytes`/`List[bytes]` <-> `str`/`List[str]`) at the input/output-level within the request/response payload
  - encode/decode function should be the same function for the conversions and it would just be calling it at different levels
  - run conversion before sending to handler
- [x] Add documentation/example

--- possible split ----
- [x] implement content type decoding/encoding (`bytes`/`List[bytes]` <-> `str`/`List[str]`) at the request/response-level within the request/response payload

--- possible split ---
- [ ] implement content type decoding/encoding (`bytes`/`List[bytes]` <-> `str`/`List[str]`) at the input/output-level within the model metadata
- [ ] implement content type decoding/encoding (`bytes`/`List[bytes]` <-> `str`/`List[str]`)  at the request/response-level within the model metadata

## Test Strategy/Script
1. Build the aissemble-open-inference-protocol project - `mvnd clean install`
2. In a new test directory, create a temp venv and pip install aissemble-open-inference-protocol-fastapi.

```
mkdir test-oip && cd test-oip
python -m venv venv
source venv/bin/activate
```
**For OTS - pip install the whl**
```
pip install {path/to}/aissemble-open-inference-protocol/aissemble-open-inference-protocol-fastapi/dist/aissemble_open_inference_protocol_fastapi-1.0.0.dev0-py3-none-any.whl --force-reinstall
```
**For final test - pip install the package (available via `test-pypi`)**
```
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ aissemble-open-inference-protocol-fastapi
```
3. Create `main.py` file with a simple request handler 
```
from typing import Optional, Any, List

from aissemble_open_inference_protocol_fastapi.aissemble_oip_fastapi import (
    AissembleOIPFastAPI,
)
[...]

Issue #24, opened 2025-06-23
Feature: Add tests for StringCodec


chang-annie:
## Description
Follow on ticket for #12. Now that we have the basic StringCodec to decode/encode between bytes/list of bytes <-> string/list of strings, we should add on to the examples/ directory. We should also include unit tests for the StringCodec. Also should reduce the complexity of utils/ `get_content_type` method as noted in [this](https://github.com/boozallen/aissemble-open-inference-protocol/pull/25#discussion_r2162594851) PR comment

## DOD
- [x] Create tests for StringCodec in `aissemble-open-inference-protocol-shared`
- [x] Simplify logic of `get_content_type` util method 
- [x] Add MLServer license and disclaimer to files using their logic

```gherkin
Feature: StringCodec encoding and decoding

  Scenario: Decode InferenceRequest with request level content_type
    Given an InferenceRequest with string content_type at the request level and payload ["hello", "world"]
    When I decode the payload
    Then the result should be a list of strings matching ["hello", "world"]

  Scenario: Decode InferenceRequest with input level content_type
    Given an InferenceRequest with string content_type at the input level and payload ["foo", "bar"]
    When I decode the payload
    Then the result should be a list of strings matching ["foo", "bar"]

  Scenario: Encode payload into InferenceResponse
    Given a payload with type string ["wombat", "llama"]
    When I encode the payload at the response level
    Then the result should be an InferenceResponse

  Scenario: Encode payload into ResponseOutput
    Given a payload with type string ["spam", "eggs"]
    When I encode the payload at the request level
    Then the result should be a ResponseOutput
```

## Test Strategy/Script
1. Pull latest version of [aissemble-open-inference-protocol](https://github.com/boozallen/aissemble-open-inference-protocol)
2. Run `mvnd clean install`
3. Verify in the logs you see that the `StringCodec` tests have successfully run
     -  ```
        [INFO] --- habushu:3.0.0:behave-bdd-test (default-behave-bdd-test) @ aissemble-open-inference-protocol-shared ---
        [INFO] Executing poetry command: poetry show behave
        [INFO] Executing poetry command: poetry add kappa-maki --group dev
        [INFO] Executing behave tests in /Users/lemuelwatts/Documents/aissemble-open-inference-protocol/aissemble-open- inference-protocol-shared/tests/features...
        [INFO] -------------------------------------------------------
        [INFO] T E S T S
        [INFO] -------------------------------------------------------
[...]

PR #25, opened 2025-06-23
[#12] implement string typing at input/output and request/response-le…

chang-annie:
…vel in payload


aissemble-open-inference-protocol-fastapi/README.md
```diff
@@ -34,7 +34,7 @@ from typing import Optional
 from aissemble_open_inference_protocol_fastapi.handlers.dataplane import (
     DataplaneHandler,
 )
-from aissemble_open_inference_protocol_fastapi.types.dataplane import (
+from aissemble_open_inference_protocol_shared.types.dataplane import (
```
chang-annie, line 37: I: moved the types/dataplane.py to the `aissemble_open_inference_protocol_shared` module so we can reuse it

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -61,7 +70,13 @@ def infer_model(
         authz_adapter, authorization, AUTH_ACTION_READ, AUTH_RESOURCE_DATA
     )
 
-    return handler.infer(model_name=model_name, payload=payload)
+    raw_request_payload = payload
+    decoded_payload = decode_inference_request(payload)
+    result = handler.infer(model_name=model_name, payload=decoded_payload)
+
+    return build_inference_response(
```
chang-annie, line 77: I: the decode logic can live in the shared utils because the act of taking any incoming `InferenceRequest` and turning it into either a Python payload or the `InferenceRequest` with decoded inputs will always be the same. The encoding the response logic is a bit tricky because it relies on some applications-specific logic built on top of the helpers. I wonder if providing the `build_inference_response` to users is overkill and if we should just return the result from the `handler.infer(...)` call here

jaebchoi (approved):

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -275,3 +297,67 @@ def server_metadata(
...
+            model_version=model_version,
+        )
+        if resp is not None:
+            resp.id = request.id
+            return resp
+
+    # No codec matched, just echo the raw data from the inputs
+    outputs = []
+    for inp in request.inputs or []:
+        ct = None
```
aaron-gary: A: Avoid short variable names.  I would just call it content_type.  Makes is a bit easier to read.

chang-annie: good point! updated few of the variable names to be more descriptive

aissemble-open-inference-protocol-shared/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Features
+
+### content_type Parameter (Work In Progress)
+The `content_type` feature lets you declare, at either request time or in model metadata, how each OIP-compliant input should be converted into a Python object.
+
+**Declaring Content Types:**
+- **Input Level**: Include a `content_type` field under each `inputs[].parameters` in the request JSON. This will apply the content type to the particular input only.
+- **Request Level**: Include a `content_type` field at the top-level `parameters` tag of the request JSON. This will decode the entire request into one a single Python object.
+
```
aaron-gary, line 27: A:  Can you include which one takes precedence when both levels are present? (if you don't have it further down)

aissemble-open-inference-protocol-fastapi/pyproject.toml
```diff
@@ -26,6 +26,7 @@ Homepage = "https://github.com/boozallen/aissemble-open-inference-protocol"
 packages = [{include = "aissemble_open_inference_protocol_fastapi", from = "src"}]
 
 [tool.poetry.dependencies]
+aissemble-open-inference-protocol-shared = {path = "../aissemble-open-inference-protocol-shared", develop = true}
```
cwoods-cpointe, line 29: c/q: Do we need to leverage Habushus `rewriteLocalPathDepsInArchives` for CI deploying?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/string.py
```diff
@@ -0,0 +1,128 @@
...
+logger = LogManager.get_instance().get_logger("StringCodec")
+
+DEFAULT_STR_CODEC = "utf-8"
+BytesStringList = Union[bytes, str]
+PayloadList = Union[BytesStringList, List[BytesStringList]]
+
+
+def encode_str(s: str) -> bytes:
+    return s.encode(DEFAULT_STR_CODEC)
+
```
aaron-gary, line 40: A:  I'm guessing if the data isn't encodable/decodable, the generic exception will be easy enough to track down, but you might consider wrapping them in a try/catch so you can give the user a concise explanation for the exception

aissemble-open-inference-protocol-fastapi/pyproject.toml
chang-annie: confirmed with you offline that it's already being set in the `ci` profile that gets run for our build

aissemble-open-inference-protocol-shared/README.md
chang-annie: updated!

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
cwoods-cpointe: Id say adding this complexity here is what we want because it takes the burden off the users and keeps the response format uniform.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -0,0 +1,251 @@
...
+PARAMETERS = "parameters"
+CONTENT_TYPE = "content_type"
+
+InputOrOutput = Union[RequestInput, ResponseOutput]
+Codec = Union[InputCodecType, RequestCodecType]
+ParameterizedObject = Union[
+    RequestInput, ResponseOutput, RequestOutput, InferenceRequest, InferenceResponse
+]
+
+logger = LogManager.get_instance().get_logger("Codec.Utils")
```
cwoods-cpointe: q: is this logger being used?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/string.py
chang-annie: added a try-except block!

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
chang-annie: nope, that was leftover from testing! just removed it. good catch!

cwoods-cpointe (approved):

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -0,0 +1,249 @@
...
+) -> Optional[str]:
+    """
+    Read the `content_type` tag from a ParameterizedObject or its metadata.
+    """
+    params = getattr(obj, PARAMETERS, None)
+    if params is not None:
+        content_type = getattr(params, CONTENT_TYPE, None)
+        if content_type:
+            return content_type
+    if metadata is not None:
```
aaron-gary, line 65: S: I'm not sure what to suggest, but the nested nature of these blocks is a little hard to follow.  Perhaps something like this would reduce the nesting a little:

```
if metadata:
    content_type = getattr(getattr(metadata, PARAMETERS, None), CONTENT_TYPE, None)
    if content_type:
        return content_type
```

chang-annie: sorry i merged my branch in before this comment got it! i'll add a note to #24 to reduce the complexity here

chang-annie: added a note to https://github.com/boozallen/aissemble-open-inference-protocol/issues/24

Issue #30, opened 2025-06-30
Feature: Add test to validate of FASTAPI REST Interface


jaebchoi:
## Description
As we have introduced examples for all endpoints for FastAPI REST Interfaces, we need integration test to make sure this example doesn't fail further down the road. We would like to automate validation of example. 
This validation covers all endpoints for REST OIP Endpoints. 

**UPDATE:** We will not introduce Integration test per investigation and use this issue to refactor dataplane handler back to abstract class/methods and build unit test based on DOD.

## DOD
- [x] Ensure users are required to implement dataplane handler for inference and model endpoints. ( infer, model_metadata, model_ready) 
- [x] Users use Default implementation for server endpoints. (server_live, server_ready, server_metadata) 
- [x] Default implementation (Server endpoints) can be overridden and used instance information. (server_live, server_ready, server_metadata) 

- [x] Ensure FASTAPI Examples fails to build if library have breaking change. **(Upon investigation with team, we think this is overkill and comes with heavy load of work due to nature of abstract method in python and lint limitation )**

## Test Strategy/Script

1. cd into  aissemble-open-inference-protocol project 
2. Build the aissemble-open-inference-protocol project - `mvnd clean install`
3. Verify Success that all Unit Test passed.

```feature

Feature: Customization of aissemble OIP Handler
    user should be able to provide custom implementation of handler logic so that they can integrate user's logic with ML model.

  Scenario: Handler implementations Doesn't have required methods.
    Given custom implementation of dataplane handler that doesn't implement required methods.
    When Handler is created
    Then Error is raised

  Scenario: Handler implementations have default server status methods.
    Given custom implementation of dataplane handler that doesn't implement server methods.
    When server status method is called.
    Then affirmative is responded


  Scenario: Handler implementations have overridden server status methods.
    Given custom implementation of dataplane handler that override implement server methods.
    When server status method is called.
    Then custom logic is reponded


```

jaebchoi:
DoD Completed @epwilkins @cwoods-cpointe

jaebchoi:
OTS Completed with @csun-cpointe

chang-annie:
test steps are passing! closing issue as complete.
- built successfully using `mvnd clean install` and confirmed all unit tests passed

<img width="769" alt="Image" src="https://github.com/user-attachments/assets/e7762ee8-d69b-4f77-a921-49d69d2ace13" />

PR #31, opened 2025-07-02
[#30] Add test to validate of FASTAPI REST Interface

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/30


aissemble-open-inference-protocol-fastapi/tests/features/steps/custom_handler_steps.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@then("Error is raised")
+def status_code_is_error(context):
+    assert isinstance(context.exception, TypeError)
+
+
+@then("affirmative is responded")
+def status_code_is_success(context):
+    assert context.response.status_code == 200
```
jaebchoi: I: I was trying with nt.ok but FastAPI test client doesn't support nt.ok? looking at the [doc](https://fastapi.tiangolo.com/tutorial/testing/) they seems to be compare status code so I went ahead with this, but open to ideas!

lemuelwatts: FWIW I was just able to convert your asserts to `nt.ok_` and it appeared to work fine

jaebchoi: ooh interesting.. what specific code have you ran? and what import? 
I was trying 
```

from shutil import nt

...

assert context.response == nt.ok_
```

Is this right code? im bit new to python and couldn't find exact import

aissemble-open-inference-protocol-fastapi/tests/features/custom_handler.feature
```diff
@@ -0,0 +1,19 @@
...
+
+  Scenario: Handler implementations doesn't have required methods
+    Given custom implementation of dataplane handler that doesn't implement required methods
+    When model method is called
+    Then Error is raised
+
+  Scenario: Handler implementations have default server status methods
+    Given custom implementation of dataplane handler that doesn't implement server methods
+    When server status method is called
+    Then affirmative is responded
```
aaron-gary, line 12: Q: does this mean that the default for an unimplemented status method is that the status is ready?  If so, I would think the status would default to unknown.

aissemble-open-inference-protocol-fastapi/tests/features/steps/custom_handler_steps.py
lemuelwatts: I used `import nose.tools as nt` (will also need `nose = ">=1.3.7"` in your `pyproject.toml`  in the `[tool.poetry.group.dev.dependencies]`) and then compared it like so

`nt.ok_(context.response.status_code == 200, f"Expected status code 200, got {context.response.status_code}")`

aaron-gary (approved):

lemuelwatts (approved):

aissemble-open-inference-protocol-fastapi/tests/features/custom_handler.feature
jaebchoi: Yes, reason why we put default behavior for server to ready to true is beacause whenever those `/health/ready` or `/health/live` hit this endpoint then server is able to serve request and healthy. so we feel that this can be same behavior for most of the server. But if they choose to go different logic, we can alway override this method and do its custom logic.


Not for the model thou because model have its different behavior in loading up their models so we keep this as unknown.

aissemble-open-inference-protocol-fastapi/tests/features/steps/custom_handler_steps.py
jaebchoi: awesome, Ill update based on this, Thanks Lemuel!

PR #38, opened 2025-07-07
[37] Add IP logging to Authz and simplify logging format

aaron-gary:


aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/xacml3_builder.py
```diff
@@ -121,5 +121,5 @@ def build_request(self, user, role, resource, action):
         request_dict = {"Request": request.to_dict()}
 
         json_string_request = json.dumps(request_dict, indent=2)
-        print(json_string_request)
+
```
aaron-gary, line 124: I: Logging/printing the XACML is not a good practice

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -59,17 +65,25 @@
 def infer_model(
     model_name,
     payload: InferenceRequest,
+    request: Request,
```
aaron-gary, line 76: I: The FastAPI request is needed to get the user's IP address

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
```diff
@@ -27,22 +27,40 @@ def _authorize_impl(
...
+        
+        self.logger.info(f"PDP decision for user: {user}, was {permit_or_deny}")
+
+
+    def log_authorize(self, token_data: dict, resource: str, action: str, user_ip: str, request_url: str):
+        authz_log_info = {
+            "user": token_data.get("sub"),
+            "resource": resource,
+            "action": action,
+            "role": token_data.get("role"),
```
epwilkins: S: I believe this will give an error if there is not a role attached to the token

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
```diff
@@ -27,22 +27,40 @@ def _authorize_impl(
         pass
 
     def authorize(
-        self, user: dict, resource: str, action: str, role: Optional[str] = None
+        self, token_data: dict, resource: str, action: str, user_ip: str, request_url: str
```
epwilkins: A: IMO it's better to pass user/role separately, to minimize the chance that someone logs the token.

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
```diff
@@ -27,22 +27,40 @@ def _authorize_impl(
...
 
-    def log_authorize(
-        self, user: dict, resource: str, action: str, role: Optional[str] = None
-    ):
-        user_for_logging = json.dumps(user, indent=2)
+        self.log_pdp_decision(token_data, pdp_decision)
+
+        return pdp_decision
+
+    def log_pdp_decision(self, token_data: dict, pdp_decision):
```
epwilkins: A: Same note here about passing the user only instead of the full token data

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/jwt_auth.py
```diff
@@ -52,7 +55,7 @@ def verify_jwt_token(authorization):
     return payload
 
 
-def authenticate_and_authorize(authz_adapter, authorization, action, resource):
+def authenticate_and_authorize(authz_context: AuthzContext):
```
epwilkins: S: Shouldn't this be written against some generic base class instead of Authz directly?

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -59,17 +65,25 @@
...
     authz_adapter: DefaultAdapter = Depends(DefaultAdapter),
-    authorization: str = Depends(security),
+    bearer_token: str = Depends(security),
 ) -> InferenceResponse:
     """
     Perform inference using the specified model and return the prediction results.
     """
-    authenticate_and_authorize(
-        authz_adapter, authorization, AUTH_ACTION_READ, AUTH_RESOURCE_DATA
+    authz_context = AuthzContext(
```
epwilkins: C: We need some way for the user to specify what flavor of context to use.  In Java I'd suggest something like a Prototype pattern, but there's probably a more "pythonic" way to do a similar thing.

epwilkins: Looks like [partials](https://docs.python.org/3/library/functools.html#functools.partial) might be that more pythonic answer. But IMO it seems more complicated for the _user_ over just simply creating a prototype object.

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -363,3 +441,28 @@ def build_inference_response(
         id=request.id,
         outputs=outputs,
     )
+
+
+def get_user_ip_from_request(request: Request):
```
epwilkins: A: might need to check that this is conventional for top-level functions, but if it is recommend marking this as "internal"
```suggestion
def _get_user_ip_from_request(request: Request):
```

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
```diff
@@ -363,3 +441,28 @@ def build_inference_response(
...
+    """
+    This method extracts the user's IP address from the request object.
+    The x-forwarded-for header is used by proxy servers to preserve
+    the actual IP of the user, so if it's present we need to extract
+    the IP from it.
```
epwilkins: A: Write for consumer, not implementer (also this is mainly just an excuse to use the new multi-line feature in Github 😄)
```suggestion
    Extracts the user's IP address from the request object, respecting the
    `x-forwarded-for` header if present.
```

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/jwt_auth.py
epwilkins: Also, consider renaming to just `authorize`

epwilkins (approved):

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
aaron-gary: It doesn't error out.  It just returns None and if it's converted to json, it replaces None with null.  I contemplated removing role, since we are going to replace it with attributes, but I want us to cross that bridge when we get there.

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
aaron-gary: Ya, good point.

- [x] Remove token from params and replace with data from the token

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
aaron-gary: - [x] Remove token from params and replace with data from the token

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/auth/auth_adapter_base.py
epwilkins: Ah interesting. I didn't realize the behavior difference between `dict.get(key)` and `dict[key]`

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
aaron-gary: Since we are using Python I think partials are appropriate.

- [x] Leverage partials to reduce redundant code

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
aaron-gary: Agree, this can be internal

- [x] Use python syntax for scope

aissemble-open-inference-protocol-fastapi/src/aissemble_open_inference_protocol_fastapi/rest/endpoints.py
aaron-gary: - [x] make code comment concise by using the suggested phrasing above

PR #39, opened 2025-07-08
[#27] Enable grpc servicer inferencing

### LINKED ISSUE #27: Feature: Enable inferencing for gRPC server
## Description
#9 created the gRPC server implementation. This ticket should focus on creating the gRPC server servicer to run OIP-compliant inferencing.  
Q: if we should have the same handler functionality between the different aissemble-oip implementations. 
A: Yes we should only expose one handler schema and handle converting all our oip requests to that handler schema

## DOD
- [x] Users are able to call gRPC server to perform inferencing
    - [x] Users should be able to define and inject their own handler for the servicer
    - 🟡 The handler is compatible with both the FastAPI and gRPC server
        - [x] Map from grpc request
        - 🟡 Map to grpc response
- [x] Create testing harness for servicer
----- split ------
- Map to grpc response
- Create gRPC example
- gRPC servicer for the other endpoints

## Test Strategy/Script
```feature
Scenario: When an inference request is sent to the gRPC servicer, it is successfully routed to the handler
  Given a handler to perform inferencing exists
  And an aissemble oip gRPC servicer exists with the handler
  And an infer request exists
  When an infer request is sent to the gRPC servicer
  Then the handler receives the request with the expected data
```

### Steps
1. Build aissemble-open-inference-api and verify all tests pass - `mvnd clean install`
2. Download the test project. It has a handler, grpc server, and grpc client that builds and sends requests - 
[oip-grpc.zip](https://github.com/user-attachments/files/21129093/oip-grpc.zip)

3. Add aissemble oip-grpc library -  `poetry add path/to/aissemble-open-inference-protocol/aissemble-open-inference-protocol-grpc/dist/aissemble_open_inference_protocol_grpc-1.0.0.dev0-py3-none-any.whl`
4. Build with `mvn clean install`
5. Start the server and wait for it to come up - `poetry run run_server`
6. Send a request using the client - `poetry run inference_request`
7. Verify the inference request returns without failing. You should see some basic information like `model_name` and `model_version`
8. The server is mapping the grpc request to what the custom handler is needing. To validate it is correct, the handler will print out what it receives. These logs are outputted by the server.
9. Verify the output matched the clients constructed request
    1. model_name="test_model"
    2. model_version="123",
    3. id="a test request",
    4. parameters={
    "my_bool_param": True
    "content_type": None
    "my_int64_param": 123,
    "my_string_param": some_param",
    "my_double_param": 1.1,
    "my_uint64_param": 123,
},
    5. inputs:
        1. name="input",
        2. datatype="INT64",
        3. shape=[1, 3],
        4. parameters={
    "content_type": str,
    "bool_param": True,
    "int64_param": 123,
    "string_param": some_param",
    "double_param": 1.1,
    "uint64_param": 123,
},
        5. data(root=[1, 2, 3]),
    6. outputs
        1.  name="output",
        2. parameters={
            "content_type": str
        }

[OIP gRPC proto](https://github.com/kserve/open-inference-protocol/blob/main/specification/protocol/open_inference_grpc.proto)
[MLServer gRPC](https://github.com/SeldonIO/MLServer/tree/master/mlserver/grpc)
[FastAPI gRPC](https://apidog.com/blog/fastapi-grpc/)

cwoods-cpointe:
Allow users to pass the handler to the grpc server and have the servicer call these handler functions. To keep the handler compatible with each oip impl we need to map the requests/responses to a common class.
Moved the handler to the shared module
Update the examples location to account for future examples like grpc.


aissemble-open-inference-protocol-grpc/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = [
     {name = "aiSSEMBLE Solutions Community", email = "aissemble@bah.com"}
 ]
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.11.4"
```
chang-annie, line 9: q: was there a feature in 3.11.4 that we needed? we should probably also update the classifier below to reflect `3.11.4`

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
```diff
@@ -60,7 +64,7 @@ def _add_terminate_signal_handlers(self):
         self.logger.info("Adding terminate signal handlers")
         loop = asyncio.get_running_loop()
 
-        for sig in HANDLED_SIGNALS:
+        for sig in signal.SIGINT, signal.SIGTERM, signal.SIGQUIT:
```
chang-annie, line 67: q: what's the benefit of removing the global var with these values

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -8,16 +8,58 @@
...
+from krausening.logging import LogManager
 
+from aissemble_open_inference_protocol_grpc.grpcInferenceService_pb2 import (
+    ModelInferResponse,
+)
 from aissemble_open_inference_protocol_grpc.grpcInferenceService_pb2_grpc import (
     GRPCInferenceServiceServicer,
 )
+from aissemble_open_inference_protocol_grpc.mappers.model_inference_reqeust_mapper import (
+    ModelInferenceReqeustMapper,
```
chang-annie: c: `ModelInferenceReqeustMapper` is spelled incorrectly (`Reqeust` instead of `Request`)

aissemble-open-inference-protocol-grpc/pyproject.toml
cwoods-cpointe: all the others are 3.11.4 so keeping it consistent. Good point on the classifier but 3.11 is for all 3.11.x versions

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
cwoods-cpointe: ah just moved it here to read easier. Think either is fine but global is probably overkill

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -8,16 +8,58 @@
...
+        try:
+            self.logger.info("Sending model inference request to the handler")
+            # Send request to handler
+            handler_response = self.handler.infer(
+                payload=inference_request,
+                model_name=request.model_name,
+                model_version=request.model_version,
+            )
+
+            # TODO convert InferenceResponse to ModelInferResponse
```
aaron-gary, line 55: Q: Am I reading this TODO right?  It looks like it's already ModelInferResponse.

chang-annie (approved):
there are 3 files with merge conflicts that must be resolved but otherwise LGTM

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_request_mapper.py
```diff
@@ -0,0 +1,113 @@
...
+            ],
+            data=self._infer_tensor_contents_to_data(model_infer_inputs.contents),
+        )
+
+    def _infer_tensor_contents_to_data(
+        self, tensor_contents: InferTensorContents
+    ) -> TensorData:
+        """
+        Maps ModelInferTensorContents to InferenceRequest compatible TensorData
+        :param tensor_contents: content to map. Comes in with formate - type_contents: [Any]
```
aaron-gary, line 71: formate --> format 

(a few more instances below as well)

aissemble-open-inference-protocol-grpc/tests/features/steps/grpc_servicer.py
```diff
@@ -0,0 +1,164 @@
...
+        context.handler.model_version,
+        "Model version did not map correctly to inference request",
+    )
+    nt.assert_equal(
+        expected_infer_request.id,
+        actual_infer_request.id,
+        "ID did not map correctly to inference request",
+    )
+    _assert_parameters(
+        expected_infer_request.parameters, actual_infer_request.parameters[0]
```
aaron-gary, line 110: Q: Are these both not lists?  Why is the expected_infer_request parameters being compared to the first element of actual_infer_request parameters?

aaron-gary (approved):

PR #46, opened 2025-07-14
[#40] finalize gRPC inferencing

### LINKED ISSUE #40: Feature: Finalize inferencing for gRPC server
## Description
#27 started the work to allow the gRPC server servicer to run OIP-compliant inferencing. Specifically, it focused on allowing the gRPC servicer to communicate with the handler (mapping from the gRPC `ModelInferRequest` to the `InferenceRequest`). This ticket should focus on mapping the handler's response back to the gRPC servicer (`InferenceResponse` to gRPC's `ModelInferResponse`)

## DOD
- [x] Map the OIP-compliant inference response to gRPC response
- [x] Correct model construction
```feature
Scenario: When the custom handler returns an inference response, gRPC servicer successfully returns the data
  Given a handler to perform inferencing exists
  And an aissemble oip gRPC servicer exists with the handler
  And the handler has model inferencing results
  When an infer response is sent to the handler
  Then the servicer's response corresponds to the handler's results
  
```
----- split ------
Create gRPC example
gRPC servicer for the other endpoints

## Test Strategy/Script
1. Build `aissemble-open-inference-protocol` and confirm all tests pass - `mvnd clean install`
2. Download test project [oip-grpc.zip](https://github.com/user-attachments/files/21240378/oip-grpc.zip) - includes handler, grpc server, grpc client that builds and sends requests
3. Add aissemble-open-inference-protocol-grpc library 
- for OTS: 
```
poetry add path/to/aissemble-open-inference-protocol/aissemble-open-inference-protocol-grpc/dist/aissemble_open_inference_protocol_grpc-1.0.0.dev0-py3-none-any.whl
```
- for Final Test: 
```
poetry source add --priority=supplemental testpypi https://test.pypi.org/simple/
poetry add aissemble_open_inference_protocol_grpc
```
4. Build test project -`mvn clean install`
5. Start the server and wait for it to come up `poetry run run_server`
6. Send a request using the client - `poetry run inference_request`
7. Since we are mapping the custom handler's OIP-compliant inference response back to the gRPC response, we can print and compare the outputs from the handler (the server logs) vs. outputs from the gRPC servicer (the client logs)
8. In the server logs, you should see the following
 ```
InferenceResponse before conversion:

model_name: test_model
model_version: 123
id: a test request
parameters:
  content_type: None
  my_uint64_param: 123
  my_bool_param: True
  my_double_param: 1.1
  my_string_param: some_param
  my_int64_param: 123
output.name: output-0
output.shape: [1, 3]
output.datatype: Datatype.INT64
output.data: [1, 2, 3]
output.parameters:
  content_type: str
```
9. In the client logs, you should see:
```
ModelInferenceResponse after conversion:

response type: <class 'grpcInferenceService_pb2.ModelInferResponse'>
model_name: test_model
model_version: 123
id: a test request
parameters:
  my_string_param: some_param (string_param)
  my_double_param: 1.1 (double_param)
  my_uint64_param: 123 (int64_param)
  my_int64_param: 123 (int64_param)
  my_bool_param: True (bool_param)
output.name: output-0
output.shape: [1, 3]
output.datatype: Datatype.INT64
output.contents (int64_contents): [1, 2, 3]
output.parameters:
  content_type: str (string_param)
```

#27

### PR DESCRIPTION
**Pull request by chang-annie**
[#40]
- add logic to map the handler's response back to the gRPC servicer (`InferenceResponse` to gRPC's `ModelInferResponse`)
- update model construction

### CODE REVIEW

File: aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py [hunk 1]
```diff
@@ -19,6 +19,12 @@
...
 )
+from aissemble_open_inference_protocol_grpc.mappers.model_inference_response_mapper import (
+    ModelInferenceResponseMapper,
+)
+from aissemble_open_inference_protocol_grpc.mappers.utils import (
+    MappingException,
+)
 
 
 class InferenceServicer(GRPCInferenceServiceServicer):
```
epwilkins: A: We should add type hints to the parameters in the methods of this class

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -52,14 +58,12 @@ def ModelInfer(self, request, context) -> ModelInferResponse:
                 model_version=request.model_version,
             )
 
-            # TODO convert InferenceResponse to ModelInferResponse
```
epwilkins, line 55: Q: I'm probably just out of the loop, but aren't both of these our classes? Why would we need to convert between them instead of just having the Handler return a compatible response object?

epwilkins: Ah I see. These are the generated gRPC classes which make interacting with the protocol easier.  We should probably rename the gRPC version of the classes so the names match and it makes it more clear why there are seemingly two copies of each. E.g. `InferenceResponse` and `GrpcInferenceResponse`

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py
```diff
@@ -0,0 +1,147 @@
...
+    Datatype.BOOL: "bool_contents",
+    Datatype.UINT8: "uint_contents",
+    Datatype.UINT16: "uint_contents",
+    Datatype.UINT32: "uint_contents",
+    Datatype.UINT64: "uint64_contents",
+    Datatype.INT8: "int_contents",
+    Datatype.INT16: "int_contents",
+    Datatype.INT32: "int_contents",
+    Datatype.INT64: "int64_contents",
+    Datatype.FP16: "bytes_contents",
```
epwilkins, line 39: Q/S: Why do 16 bit floating points translate to bytes?

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py
```diff
@@ -0,0 +1,147 @@
...
+    def _inference_response_to_params(
+        self, parameters: Parameters
+    ) -> Mapping[str, InferParameter]:
+        """
+        Maps inference response parameters to model parameters
+        :param parameters: the InferenceResponse Parameters object in format - 'key': value
+        :return: the model inference parameters in format - 'key': type : value
+        """
+        param_dict = parameters.model_dump(by_alias=True)
+        params = {}
```
epwilkins: A: Because of all of the overloaded terms clashing here, it makes this a bit difficult to follow without digging through the rest of the code. Recommend making these variable names really explicit to help out. e.g. `converted_model_params`, `model_param_name`, `grpc_type_key`, etc.

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py
```diff
@@ -0,0 +1,147 @@
...
+        contents = self._create_field_mapping(tensor_contents, datatype)
+        return InferTensorContents(**contents)
+
+    def _inference_response_to_params(
+        self, parameters: Parameters
+    ) -> Mapping[str, InferParameter]:
+        """
+        Maps inference response parameters to model parameters
+        :param parameters: the InferenceResponse Parameters object in format - 'key': value
+        :return: the model inference parameters in format - 'key': type : value
```
epwilkins: A: The proposed return format doesn't really make sense to me. We probably just need to be a bit wordier, but `'key' : type : value` makes me think it's maybe a map of maps which it is only very conceptually. The docs for `_create_field_mapping` for instance, read much more clearly

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/utils.py
```diff
@@ -0,0 +1,30 @@
...
+# %%
+# Copyright (C) 2024 Booz Allen Hamilton Inc.
+# %%
+# This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+# #L%
+###
+from typing import Mapping
+
+
+class MapperUtils:
```
epwilkins: A: Probably more pythonic to just let these functions live at the root. Then use with

```python
import ...mappers.utils

utils.merge_infer_parameters
```
and if you needed to deconflict util classes, you can `import utils as mapper_utils`

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
epwilkins: Also, I know this was not done in this PR, but the Protobuf [Style Guide](https://protobuf.dev/programming-guides/style/) recommends treating acronyms as single words (`GrpcInferenceService`) and naming proto files in lower snake case (`grpc_inference_service.proto`)

epwilkins (approved):

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py
chang-annie: there's no native `fp16` data type but we can use `fp16 `data by first representing it as `bytes` field within the protobuf message and then deserialize/extract the `bytes` field downstream

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
chang-annie: spoke offline with @epwilkins and agreed to keep current gRPC class names to keep them in line with kserve and mlserver implementation

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
chang-annie: regenerated and replaced the protobuf files with the correct naming style

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py
chang-annie: cleaned up the docstring & updated the variable names to be more explicit

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/mappers/utils.py
chang-annie: moved the functions to the root and updated imports and usage accordingly

PR #66, opened 2025-07-31
[#52] build_inference_response shared function

### LINKED ISSUE #52: Feature: Inference response encoding is uphled by both request and handler content types
## Description
Response content type encoding should be supported by two flows. One is the handler response has content type. The other is the request has the content type. We should support encoding the response for both of these.

**The handlers response**
In some cases the model or handler can only support specific data types. Users can define this by setting the response content type. If the handler response has defined the parameter content type, we should encode the response to this regardless of what the request asks for. 

**The request**
The request can ask for the data as a specific type. They do this by setting the request outputs content-type. It can be defined at an individual output level or at the entire response level. The individual outputs should match the response outputs by `name`.

## DOD
- [x] The model's inference handler content type is upheld and used to encoded the results.
- [x] The user's request content type is upheld and used to encoded the results if the response handler doesn't already have a type defined.
  - [x] Users can leverage content at a request level
  - [x] Users can leverage content at a output level
- [x] Documentation is created showing how to correctly use the codec functionality
    ~- [ ] Possible example can be created if it provides users value in how to leverage it~
- [x] Unit tests are in place to cover the request and response encoding/decoding

## Test Strategy/Script
1. Pull the latest version of [aissemble-open-inference-protocol](https://github.com/boozallen/aissemble-open-inference-protocol) and run `mvn clean install` in the root of the project
2. Download example project ([aissemble-oip-response-encoding.zip](https://github.com/user-attachments/files/21553730/aissemble-oip-response-encoding.zip)) which contains a handler and server to accept and process requests
3. `cd` into the project folder and add the fastapi module to your project with the following command:  
    - ```
      poetry add aissemble-open-inference-protocol-fastapi
       ```
4. Start the server with the following command: `poetry run fastapi dev ./src/aissemble-oip-fastapi-inference/main.py`
5. Execute the following `curl` commands and verify the output you receive is what we are expecting:
### Scenario 1: Handler Output Content-Type Takes Precedence
```
curl -X POST "http://127.0.0.1:8000/v2/models/handler-precedence-model/infer" \
  -H "Content-Type: application/json" \
  -d '{
    "id": "test-handler-precedence",
    "inputs": [{"name": "input-0", "shape": [3], "datatype": "FP32", "data": [1,2,3]}],
    "outputs": [
      {
        "name": "test_output",
        "parameters": {
          "content_type": "application/json"
        }
      }
    ],
    "parameters": {
      "content_type": "application/json"
    }
  }'
```
**Expected Result:** The output should have content_type: "application/handler-override" (handler wins over request)


### Scenario 2: Request Output Content-Type Used When Handler Omits

```
curl -X POST "http://127.0.0.1:8000/v2/models/request-fallback-model/infer" \
  -H "Content-Type: application/json" \
  -d '{
    "id": "test-request-fallback",
    "inputs": [{"name": "input-0", "shape": [3], "datatype": "FP32", "data": [1,2,3]}],
    "outputs": [
      {
        "name": "test_output",
        "parameters": {
          "content_type": "str"
        }
      }
    ]
  }'
```

**Expected Result:** The output should have content_type: "str"(request output content-type is used)

### Scenario 3: Request-Level Content-Type Applies to All Outputs
```
curl -X POST "http://127.0.0.1:8000/v2/models/request-level-model/infer" \
  -H "Content-Type: application/json" \
  -d '{
    "id": "test-request-level",
    "inputs": [{"name": "input-0", "shape": [3], "datatype": "FP32", "data": [1,2,3]}],
    "outputs": [
      {
        "name": "output1"
      },
      {
        "name": "output2"
      },
      {
        "name": "output3"
      }
    ],
    "parameters": {
      "content_type": "str"
    }
  }'
```
**Expected Result:** All three outputs should have content_type: "str" (request-level content-type applies to all)

### Scenario 4: All outputs produced by model are returned with no output requested and Request-Level Content-Type Applies to All Outputs
```
curl -X POST "http://127.0.0.1:8000/v2/models/request-level-model/infer" \
  -H "Content-Type: application/json" \
  -d '{
    "id": "test-request-level",
    "inputs": [{"name": "input-0", "shape": [3], "datatype": "FP32", "data": [1,2,3]}],
    "outputs": [],
    "parameters": {
      "content_type": "str"
    }
  }'
```
**Expected Result:** Three outputs from the model are returned (e.g. "name":"output1-from-model") even though we didn't specify any outputs

A clear and concise description of any alternative solutions or features you've considered.
Add any other context, links, or screenshots about the feature request here.

lemuelwatts:
#52

### DIFF
--- a/aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
+++ b/aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
@@ -179,6 +179,41 @@ def encode_inference_response(
     )
 
 
+def build_inference_response(
+    model_name: str,
+    request: InferenceRequest,
+    result: InferenceResponse,
+    model_version: Optional[str] = None,
+) -> InferenceResponse:
+    """
+    Construct an InferenceResponse by encoding a handler's raw Python result according to content_type
+    1. Handler output content_type takes precedence.
+    2. Try per-output content_type (if request.outputs is set).
+    3. Fallback to a request-level content_type (if request.parameters.content_type is set).
+    4. Otherwise, fallback to None content_type.
+    """
+    # Import here to avoid circular import
+    from aissemble_open_inference_protocol_shared.codecs.strategies.output_processing_strategy import (
+        ProcessAllOutputsStrategy,
+        ProcessRequestedOutputsStrategy,
+    )
+
+    request_outputs = getattr(request, "outputs", None) or []
+    if not request_outputs:
+        strategy = ProcessAllOutputsStrategy(request, result)
+    else:
+        strategy = ProcessRequestedOutputsStrategy(request, result)
+
+    outputs = strategy.process_outputs()
+
+    return InferenceResponse(
+        model_name=model_name,
+        model_version=model_version,
+        parameters=getattr(result, "parameters", None),
+        outputs=outputs,
+    )
+
+
 class SingleTensorRequestCodec(RequestCodec):
     """
     Base class for request-level codecs that wrap a single tensor.
--- a/aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
+++ b/aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
@@ -1,0 +1,98 @@
+from behave import given, when, then
+import nose.tools as nt
+from aissemble_open_inference_protocol_shared.types.dataplane import (
+    InferenceRequest,
+    InferenceResponse,
+    ResponseOutput,
+    RequestOutput,
+    Parameters,
+    TensorData,
+    Datatype,
+)
+from aissemble_open_inference_protocol_shared.codecs.utils import (
+    build_inference_response,
+)
+
+
+@given('a handler returns output "{output_name}" with content type "{content_type}"')
+def step_given_handler_output_with_content_type(context, output_name, content_type):
+    output = ResponseOutput(
+        name=output_name,
+        shape=[1],
+        datatype=Datatype.BYTES,
+        parameters=Parameters(content_type=content_type),
+        data=TensorData(root="handler-data"),
+    )
+    context.handler_response = InferenceResponse(
+        model_name="test-model",
+        outputs=[output],
+    )
+
+
+@given('a handler returns output "{output_name}" without content type')
+def step_given_handler_output_without_content_type(context, output_name):
+    output = ResponseOutput(
+        name=output_name,
+        shape=[1],
+        datatype=Datatype.BYTES,
+        parameters=Parameters(),
+        data=TensorData(root="handler-data"),
+    )
+    context.handler_response = InferenceResponse(
+        model_name="test-model",
+        outputs=[output],
+    )
+
+
+@given('a request asks for output "{output_name}" with content type "{content_type}"')
+def step_given_request_output_with_content_type(context, output_name, content_type):
+    output = RequestOutput(
+        name=output_name,
+        parameters=Parameters(content_type=content_type),
+    )
+    context.request_outputs = [output]
+
+
+@given('a request asks for content type "{content_type}" at request level')
+def step_given_request_level_content_type(context, content_type):
+    context.request_parameters = Parameters(content_type=content_type)
+
+
+@when("the response is processed")
+def step_when_response_processed(context):
+    request = InferenceRequest(
+        inputs=[],
+        outputs=getattr(context, "request_outputs", None),
+        parameters=getattr(context, "request_parameters", None),
+    )
+    context.response = build_inference_response(
+        model_name="test-model",
+        request=request,
+        result=context.handler_response,
+    )
+
+
+@then('output "{output_name}" should use content type "{expected_content_type}"')
+def step_then_output_should_use_content_type(
+    context, output_name, expected_content_type
+):
+    found = False
+    for output in context.response.outputs:
+        if output.name == output_name:
+            nt.eq_(
+                output.parameters.content_type,
+                expected_content_type,
+                f'Output "{output_name}" content_type was "{output.parameters.content_type}", expected "{expected_content_type}"',
+            )
+            found = True
+    nt.ok_(found, f"Output {output_name} not found in response")
+
+
+@then('all outputs should use content type "{expected_content_type}"')
+def step_then_all_outputs_should_use_content_type(context, expected_content_type):
+    for output in context.response.outputs:
+        nt.eq_(
+            output.parameters.content_type,
+            expected_content_type,
+            f'Output "{output.name}" content_type was "{output.parameters.content_type}", expected "{expected_content_type}"',
+        )
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ target/
 .venv/
 __pycache__/
 .python-version
+.vscode/
--- a/aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference/poetry.lock
+++ b/aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference/poetry.lock
@@ -955,6 +955,7 @@ optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 markers = "python_version == \"3.9\""
+markers = "python_version == \"3.9\""
 files = [
     {file = "keras-3.10.0-py3-none-any.whl", hash = "sha256:c095a6bf90cd50defadf73d4859ff794fad76b775357ef7bd1dbf96388dae7d3"},
     {file = "keras-3.10.0.tar.gz", hash = "sha256:6e9100bf66eaf6de4b7f288d34ef9bb8b5dcdd62f42c64cfd910226bb34ad2d2"},
@@ -972,15 +973,15 @@ rich = "*"
 
 [[package]]
 name = "keras"
-version = "3.11.0"
+version = "3.11.1"
 description = "Multi-backend Keras"
 optional = false
 python-versions = ">=3.10"
 groups = ["main"]
 markers = "python_version >= \"3.10\""
 files = [
-    {file = "keras-3.11.0-py3-none-any.whl", hash = "sha256:28babe8dc69370a2b7c3bc385be6c1f22d702b00b3c47cae62559b6f80d6073a"},
-    {file = "keras-3.11.0.tar.gz", hash = "sha256:f5dfeaf4fcaea180e032f7c1e373f1868961e2940dcfcaaf9a5b711baf41bd60"},
+    {file = "keras-3.11.1-py3-none-any.whl", hash = "sha256:641dc7535fb737fce7841363683af31cb4dc5f1b1b5d961d3f06982eb29d8c7a"},
+    {file = "keras-3.11.1.tar.gz", hash = "sha256:7a27f384467fa8d0b0281665b52efd6bd948f20854099e35929786ce44d847f0"},
 ]
 
 [package.dependencies]
@@ -1949,11 +1950,14 @@ files = [
 [[package]]
 name = "sentry-sdk"
 version = "2.34.1"
+version = "2.34.1"
 description = "Python client for Sentry (https://sentry.io)"
 optional = false
 python-versions = ">=3.6"
 groups = ["main"]
 files = [
+    {file = "sentry_sdk-2.34.1-py2.py3-none-any.whl", hash = "sha256:b7a072e1cdc5abc48101d5146e1ae680fa81fe886d8d95aaa25a0b450c818d32"},
+    {file = "sentry_sdk-2.34.1.tar.gz", hash = "sha256:69274eb8c5c38562a544c3e9f68b5be0a43be4b697f5fd385bf98e4fbe672687"},
     {file = "sentry_sdk-2.34.1-py2.py3-none-any.whl", hash = "sha256:b7a072e1cdc5abc48101d5146e1ae680fa81fe886d8d95aaa25a0b450c818d32"},
     {file = "sentry_sdk-2.34.1.tar.gz", hash = "sha256:69274eb8c5c38562a544c3e9f68b5be0a43be4b697f5fd385bf98e4fbe672687"},
 ]
--- a/aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/poetry.lock
+++ b/aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/poetry.lock
@@ -1053,6 +1053,7 @@ optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 markers = "python_version == \"3.9\""
+markers = "python_version == \"3.9\""
 files = [
     {file = "keras-3.10.0-py3-none-any.whl", hash = "sha256:c095a6bf90cd50defadf73d4859ff794fad76b775357ef7bd1dbf96388dae7d3"},
     {file = "keras-3.10.0.tar.gz", hash = "sha256:6e9100bf66eaf6de4b7f288d34ef9bb8b5dcdd62f42c64cfd910226bb34ad2d2"},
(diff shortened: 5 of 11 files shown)


aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+    request: InferenceRequest,
+    result: InferenceResponse,
+    model_version: Optional[str] = None,
+) -> InferenceResponse:
+    """
+    Construct an InferenceResponse by encoding a handler's raw Python result according to content_type
+    1. Handler output content_type takes precedence.
+    2. Try per-output content_type (if request.outputs is set).
+    3. Fallback to a request-level content_type (if request.parameters.content_type is set).
+    4. Otherwise, fallback to string content_type.
```
cwoods-cpointe: q/c: why fall back to a content type they didnt set? If they didnt request a type and the handler didnt require one, I think we shouldnt assume and encode

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+            if (
+                handler_output.parameters
+                and hasattr(handler_output.parameters, "content_type")
+                and handler_output.parameters.content_type
+            ):
+                outputs.append(handler_output)
+                continue
+
+            output = encode_response_output(payload, request_output)
+            if output:
```
cwoods-cpointe: a: this if block is duplicated so can be a function

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+
+                params.content_type = _determine_content_type(
+                    params, request_output, request_content_type
+                )
+                output.parameters = params
+                outputs.append(output)
+
+    # If request.outputs is not set, iterate over handler outputs
+    else:
+        for handler_output in result.outputs:
```
cwoods-cpointe: c: The handler output content-type should take president over the request.outputs content type. The way this if else is set up, the results contenttype is ignored if request has some

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+                elif not params:
+                    params = Parameters()
+
+                params.content_type = _determine_content_type(
+                    params, request_output, request_content_type
+                )
+                output.parameters = params
+                outputs.append(output)
+
+    if outputs:
```
cwoods-cpointe: q: is it possible some of the request.outputs have content-type but not all of them? Currently if the request level has a type and one request.output has a type but the other outputs dont, only the one output will be encoded.
Also calls into question if only some of the results have content type while others dont. Currently only those results will be encoded and it will ignore all request.outputs

cwoods-cpointe: From the docs I would assume each output is treated individually and goes through the precedence order.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+    2. Try per-output content_type (if request.outputs is set).
+    3. Fallback to a request-level content_type (if request.parameters.content_type is set).
+    4. Otherwise, fallback to string content_type.
+    """
+    outputs: List[ResponseOutput] = []
+    request_content_type = get_content_type(request)
+    result_outputs_map = {res_output.name: res_output for res_output in result.outputs}
+
+    # Per-output content_type
+    if request.outputs:
```
cwoods-cpointe: c: These if/else are setting the params.content-type but not performing the encoding.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,137 @@ def encode_inference_response(
...
+    return res_output.data.root if hasattr(res_output.data, "root") else res_output.data
+
+
+def _determine_content_type(
+    handler_params: Optional[Parameters],
+    request_output: RequestOutput,
+    request_content_type: Optional[str],
+) -> str:
+    """Determines the content type based on a priority order."""
+    handler_content_type = get_content_type(handler_params)
```
cwoods-cpointe: s: `get_content_type` does not account for `handler_params` being None but here it is optional

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
cwoods-cpointe: I see now we are calling `encode_response_output`

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,77 @@ def encode_inference_response(
...
+    1. Handler output content_type takes precedence.
+    2. Try per-output content_type (if request.outputs is set).
+    3. Fallback to a request-level content_type (if request.parameters.content_type is set).
+    4. Otherwise, fallback to None content_type.
+    """
+    outputs: List[ResponseOutput] = []
+    request_content_type = get_content_type(request)
+    result_outputs_map = {res_output.name: res_output for res_output in result.outputs}
+
+    for request_output in getattr(request, "outputs", None) or []:
```
cwoods-cpointe: q: What happens if the handler response output isnt in the request outputs?

lemuelwatts: If the handler response output isn't in the request outputs, it will be ignored and not included in the final response. Only outputs requested by the client are returned.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,77 @@ def encode_inference_response(
...
+def _get_payload(res_output: ResponseOutput) -> Any:
+    """Extracts payload from a ResponseOutput."""
+    return res_output.data.root if hasattr(res_output.data, "root") else res_output.data
+
+
+def _determine_content_type(
+    handler_params: Parameters,
+    request_output: RequestOutput,
+    request_content_type: str,
+) -> str:
```
cwoods-cpointe:
```suggestion
) -> Optional[str]:
```

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,77 @@ def encode_inference_response(
...
+        )
+        payload = _get_payload(handler_output) if handler_output else None
+        if content_type and payload is not None:
+            # Create a patched RequestOutput with the determined content_type
+            encoded = encode_response_output(payload, request_output)
+            # The encoded ResponseOutput should already have the correct content_type
+            if encoded:
+                outputs.append(encoded)
+                continue
+        if handler_output and isinstance(handler_output, ResponseOutput):
```
cwoods-cpointe: q: Is it ever possible for handler_output to not be an instance of ResponseOutput? isinstance can have some poor performance

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,77 @@ def encode_inference_response(
...
+            outputs.append(handler_output)
+        elif isinstance(request_output, ResponseOutput):
+            outputs.append(request_output)
```
cwoods-cpointe: q: why are we appending the request outputs?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,77 @@ def encode_inference_response(
...
+        content_type = _determine_content_type(
+            handler_params, request_output, request_content_type
+        )
+        payload = _get_payload(handler_output) if handler_output else None
+        if content_type and payload is not None:
+            # Create a patched RequestOutput with the determined content_type
+            encoded = encode_response_output(payload, request_output)
+            # The encoded ResponseOutput should already have the correct content_type
+            if encoded:
+                outputs.append(encoded)
```
cwoods-cpointe: q: does this encoded output need to contain the param content-type? I think if the type came from the request and not the response then `encoded` will not have it

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: I was doing that as kind of a fallback for when there is no matching handler output for a requested output. this way the client is returned all outputs that are requested even if the handler didn't produce a result for them

aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
```diff
@@ -0,0 +1,91 @@
...
+
+@given('a request asks for content type "{content_type}" at request level')
+def step_given_request_level_content_type(context, content_type):
+    context.request_parameters = Parameters(content_type=content_type)
+
+
+@when("the response is processed")
+def step_when_response_processed(context):
+    request = InferenceRequest(
+        model_name="test-model",
```
cwoods-cpointe: a: `model_name` is not attribute of `InferenceRequest`

aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@then('output "{output_name}" should use content type "{expected_content_type}"')
+def step_then_output_should_use_content_type(
+    context, output_name, expected_content_type
+):
+    found = False
+    for output in context.response.outputs:
+        if output.name == output_name:
+            nt.eq_(output.parameters.content_type, expected_content_type)
```
cwoods-cpointe: a: Add error message to the content-type checks

aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
lemuelwatts: good catch. removed model_name

aissemble-open-inference-protocol-shared/tests/features/steps/response_encoding_steps.py
lemuelwatts: added error message for both content-type checks

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: ah true i think that's definitely possible. i'll add a check here to see if `encoded` has the content_type and if not i'll add it

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: yeah true I think it's probably overly cautious to check for the instance of handler_output since it's gotta be a ResponseOutput. didn't realize the performance hit of `isinstance`!

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: fixed!

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
cwoods-cpointe: From the [oip docs](https://kserve.github.io/website/latest/modelserving/data_plane/v2_protocol/#inference-request-json-object):
"outputs" :... Optional, if not specified all outputs produced by the model will be returned using default $request_output settings.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
cwoods-cpointe: Sounds good. Then just the comment above about isinstance

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: Updated logic so if no outputs are specified by the request then all model outputs are returned

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,102 @@ def encode_inference_response(
...
+    """
+    outputs: List[ResponseOutput] = []
+    request_content_type = get_content_type(request)
+    result_outputs_map = {res_output.name: res_output for res_output in result.outputs}
+
+    # Get request outputs, defaulting to empty list if None
+    request_outputs = getattr(request, "outputs", None) or []
+
+    # If no outputs were specified in the request, include all outputs from the handler result
+    if not request_outputs:
```
cwoods-cpointe, line 202: s: I think we still need to call `encode_response_output` and encode these outputs

lemuelwatts: ah good catch. updated to attempt to encode the outputs and if that fails then just add the output

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
```diff
@@ -179,6 +179,109 @@ def encode_inference_response(
...
+
+    # Get request outputs, defaulting to empty list if None
+    request_outputs = getattr(request, "outputs", None) or []
+
+    # If no outputs were specified in the request, encode all outputs from the handler result
+    if not request_outputs:
+        outputs = []
+        for handler_output in result.outputs:
+            # Create a RequestOutput for encoding
+            request_output = RequestOutput(name=handler_output.name)
```
aaron-gary: I recommend exploring some programming patterns to simplify the nested if() blocks.  I think the logic is good, it's just a bit complex.  One pattern to look into is the strategy pattern.  As long as you can provide the context to the strategy class that covers all the criteria (preference schedule/precedence chain) you should be able to simplify the logic, or at least abstract it away so it's easy to read in these methods.

aaron-gary (approved):

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/utils.py
lemuelwatts: Refactored the code into a strategy which lives at `aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/strategies/output_processing_strategy.py` and separates the logic out into 2 separate strategies based on whether the request is asking for any outputs or not (we supply all outputs from the handler if no outputs are requested)

Issue #58, opened 2025-07-22
Feature: Standardize user interaction with aissemble oip compliant inferencing servers for a more consistent and seamless experience


chang-annie:
## Description
We should first determine what the expected user experience will be. Based on the expected flow, we can identify and standardize the key similarities between our different solutions (gRPC, fastapi, kserve, etc). Aligning on the UX and flow will be crucial since the solution will drive consistency across future interface implementations

We could create an abstract base class that all solutions must implement. The abstract should take in a handler, security adapter and initialize the Krausening configuration files. 

We should revisit how we want to structure our configuration/properties files. Ideally, we'd take a top-down approach where we can define shared properties at a higher level, with technology-specific configuration nested underneath. This configuration could reside within `aissemble-open-inference-protocol` or potentially be part of aissemble universal config. This is TBD.

## DOD
- [x] Identify desired user experience for interacting with aissembl-OIP inferencing servers
    - Simple and easy
    - Common infernecing functionality (handlers)
    - Common authz functionality (adapters)
    - Customization of the servers
    - How users start/stop the server
- [x] The different solutions are standardized in their users interface
    - [x] Exceptions the standard approach are well documented
- [x] Configuration is defined consistently across implementations
- [x] Default handler functionality is consistent and valid for each solution
- [x] Servers are exposed and customizable
- [x] Update examples accordingly
- [x] Update documentation accordingly 

---- Potential Split ----
- [ ] Desired configuration for the solutions is identified
- [ ] The solutions share a standard configuration implementation 

## Test Strategy/Script
1. Download the skeleton project [standardize](https://github.com/user-attachments/files/21584477/standardize.zip) that contains simple habushu apps and the handler. Mostly taken from the examples project but made easier to execute
2. Navigate to each of the `main.py` implementations and initialize the server. Update the `TODO` with
```
standardize/fastapi-test/src/fastapi_test/main.py - 
fastapi_oip = AissembleOIPFastAPI(MyHandler)
origins = [
    "http://localhost",
    "http://localhost:3000",
]
fastapi_oip.server.add_middleware(
    CORSMiddleware,
    allow_origins=origins,
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)
await fastapi_oip.start()
----------
standardize/grpc-test/src/grpc_test/main.py - 
grpc = AissembleOIPgRPC(MyHandler())
grpc.server.add_insecure_port(
    address="0.0.0.0:8095",
)
[...]

PR #73, opened 2025-08-04
[#58] Created standard aissembleOIP imp and config file

cwoods-cpointe:
Going forward all implementations will use a similar abc that will uniform user experience. This updated fastapi, grpc, and kserve impls as well as put all current configuration in a common config file


aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
```diff
@@ -30,51 +31,47 @@
...
     ):
-        if inspect.isclass(handler):
-            handler = handler()
-        self.handler = handler
-        self.adapter = adapter
-        self.grpc_config = GrpcConfig(grpc_properties)
+        super().__init__(handler, adapter)
+        if inspect.isclass(self.handler):
+            self.handler = self.handler()
+        self.server = self.create_server()
```
cwoods-cpointe, line 46: I: Exposing the grpc server lets downstream apps modify the server

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
```diff
@@ -38,3 +41,7 @@ def predict(
         )
 
         return infer_response
+
+    async def start(self):
```
cwoods-cpointe, line 48: i: Our Kserve instance is not fully flushed out so not able to create a start function yet

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/aissemble_oip.py
```diff
@@ -0,0 +1,40 @@
...
+
+class AissembleOIP(ABC):
+    """
+    Abstract class for all aiSSEMBLE Open Inference Protocol solutions. Defines required standardization.
+    """
+
+    def __init__(
+        self,
+        handler: Union[DataplaneHandler, Type[DataplaneHandler]],
+        adapter: Optional[Union[AuthAdapterBase, Type[AuthAdapterBase]]],
```
cwoods-cpointe: i: Fastapi dep injection requires these to be the class object but grpc and kserve require it to be the instantiated object.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/auth/jwt_auth.py
```diff
@@ -33,6 +31,7 @@ def verify_jwt_token(authorization):
                 detail="Invalid or missing Authorization header",
             )
         try:
+            config = OIPConfig()
```
cwoods-cpointe, line 34: i: The configs were being initialized to early and being cached to None. Moved the init to where the script needs the configs

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/config/oip_config.py
```diff
@@ -19,11 +19,37 @@ class OIPConfig:
     DEFAULT_ALGORITHM = "HS256"
```
cwoods-cpointe, line 19: i: Moved the configs to one common location. The impl of this may change as we decided on a cleaner solution

aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-auth/src/aissemble-oip-fastapi-auth/main.py
```diff
@@ -25,7 +25,7 @@
 from .generate_simple_jwt import (
     create_simple_jwt_token,
 )
-from aissemble_open_inference_protocol_shared.handlers.dataplane import (
+from aissemble_open_inference_protocol_shared.handlers.dataplane_handler import (
```
epwilkins: A: Just as food for thought, but I think it would be more natural/"pythonic" for all of the provided handlers to live in the `handlers` module directly instead of all being isolated to their own.  Probably not a big enough deal to actually refactor everything, but curious if we made this decision intentionally.

aissemble-open-inference-protocol-fastapi/pyproject.toml
```diff
@@ -33,6 +33,7 @@ PyJWT="^2.10.1"
 krausening = ">=20"
 requests = ">=2.32.4"
 nose = "^1.3.7"
+uvicorn = "^0.35.0"
```
epwilkins: A: Unless there's a compelling reason (like advisement from library docs), we should stick to `>=` constraints. Also, is it correct that `nose` is listed as a direct dependency instead of under dev dependencies?

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
```diff
@@ -30,51 +31,47 @@
...
+        self, handler=DefaultHandler, adapter: AuthAdapterBase = DefaultAdapter()
     ):
-        if inspect.isclass(handler):
-            handler = handler()
-        self.handler = handler
-        self.adapter = adapter
-        self.grpc_config = GrpcConfig(grpc_properties)
+        super().__init__(handler, adapter)
+        if inspect.isclass(self.handler):
+            self.handler = self.handler()
```
epwilkins: Q: This logic confuses me now. I understand the original to be a check for whether the passed `handler` parameter is an instantiable class, and if so, calling the instantiation method of that class to create a new instance to be stored in `self.handler`. But why would `self.handler` be a class instead of an instance? Perhaps further, why was that necessary in the original logic anyway? Could we not insist an instance is passed in instead of a class? I don't see a benefit to lazily instantiating the handler.

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
```diff
@@ -30,51 +31,47 @@
...
+        self.server = self.create_server()
 
     async def start(self):
         # Add signal handlers to shut down gracefully
         self._add_terminate_signal_handlers()
 
-        self._create_server()
+        # Create server if not already created
+        if not self.server:
```
epwilkins, line 53: Q: Do we ever expect this to be true? And if so, does that mean we need some synchronization around `self.server`/`create_server`?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/config/oip_config.py
```diff
@@ -19,11 +19,37 @@ class OIPConfig:
...
         self.properties = PropertyManager.get_instance().get_properties(
             "oip.properties"
         )
 
+    #########################
+    # Authorization
+    #########################
+
+    @property
+    def grpc_auth_enabled(self) -> bool:
```
epwilkins: Q/A: Is there a use case to differentiate by technology instead of
```suggestion
    def auth_enabled(self) -> bool:
```

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/aissemble_oip.py
```diff
@@ -0,0 +1,40 @@
...
+from aissemble_open_inference_protocol_shared.auth.auth_adapter_base import (
+    AuthAdapterBase,
+)
+from aissemble_open_inference_protocol_shared.config.oip_config import OIPConfig
+from aissemble_open_inference_protocol_shared.handlers.dataplane_handler import (
+    DataplaneHandler,
+)
+
+
+class AissembleOIP(ABC):
```
epwilkins: A:
```suggestion
class AissembleOIPService(ABC):
```

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
epwilkins: I see the comment on the base class now, but it still seems to me that the tech-specific impl could insist on an instantiated object, even if the base class _can_ accept a class.

epwilkins (approved):

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
cwoods-cpointe: Yeah I think thats a fair point! I can update

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/aissemble_oip_grpc.py
cwoods-cpointe: Its a pretty odd edge case where the downstream creates the AissembleOIPgRPC() instance, overrides the server to None, doesnt create their own, then tries to start. I chose to just recreate it as the default server but could also throw exception

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/config/oip_config.py
cwoods-cpointe: Just talked with the team and we agreed we should keep this consistent across impls. Going to set it as `auth_enabled` default `true` and have fastapi also leverage it

aissemble-open-inference-protocol-fastapi/pyproject.toml
cwoods-cpointe: Yeah Ill update uvicorn constraint. nose should be under dev, good catch. Also I think requests can be removed

Issue #65, opened 2025-07-29
Feature: Add validation and error handling for OIP handler requests/responses


chang-annie:
## Description
In #62, we found that when the gRPC inference response mapper fails to serialize the response output, we get a non-descriptive error. We should improve logging and introduce more comprehensive error handling to help users in identifying/resolving issues. We should also add validation to the handler's request/response objects to confirm that they are OIP-compliant.

Sample error message:
```
File "/blah/.venv/lib/python3.12/site-packages/aissemble_open_inference_protocol_grpc/mappers/model_inference_response_mapper.py", line 106, in _converter_from_types
    return InferTensorContents(**contents)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: must be real number, not list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
 ...
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/blah/.venv/lib/python3.12/site-packages/aissemble_open_inference_protocol_grpc/inference_servicer.py", line 90, in ModelInfer
    raise MappingException("Failed to serialize inference response!")
aissemble_open_inference_protocol_grpc.mappers.utils.MappingException: Failed to serialize inference response!
```

## DOD
- [x] Validate downstream user request and response objects are OIP-compliant
  - [x] Add error handling when gRPC inference response mapper is unable to serialize response output data or if the data is in an unexpected format
- [x] Update implementing a handler section of READMEs  
- [x] Confirm examples in docs are OIP compliant

--- follow on ----
- [ ] clean up exceptions in gRPC server servicer
- [ ] handle dynamic dimensions (`-1` given in `shape`)

## Test Strategy/Script
1. Confirm build succeeds and all unit tests are passing (for OTS: `mvnd clean install`, for Final Test: [CI build](https://github.com/boozallen/aissemble-open-inference-protocol/actions/runs/16891063279)
2. Download test project [oip-grpc.zip](https://github.com/user-attachments/files/21723145/oip-grpc.zip) - includes handler, grpc server, grpc client that builds and sends requests
3. Add `aissemble-open-inference-protocol-grpc` library 
- for OTS: 
```
poetry add path/to/aissemble-open-inference-protocol/aissemble-open-inference-protocol-grpc/dist/aissemble_open_inference_protocol_grpc-1.0.0.dev0-py3-none-any.whl
```
[...]

PR #75, opened 2025-08-08
[#65] add validation for request/response objects

chang-annie:
#65 

- added validation for shape and datatype of request and response objects
- cleaned up READMEs 
- updated examples to follow new validations


aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -61,10 +61,20 @@ def ModelInfer(
             inference_request = model_inference_request_mapper.to_inference_request(
                 request
             )
-        except Exception:
-            context.set_code(grpc.StatusCode.INTERNAL)
-            context.set_details("Internal Server Error!")
-            raise MappingException("Failed to serialize model inference request!")
+        except Exception as e:
```
chang-annie, line 66: I: updated all exceptions to call [ServicerContext.abort](https://grpc.github.io/grpc/python/grpc.html#grpc.ServicerContext.abort), which will raise an exception and terminate the RPC with non-OK status

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -126,3 +155,106 @@ class ServerMetadataResponse(BaseModel):
...
+        Datatype.INT32: lambda x: isinstance(x, int),
+        Datatype.INT64: lambda x: isinstance(x, int),
+        Datatype.FP16: lambda x: isinstance(x, float),
+        Datatype.FP32: lambda x: isinstance(x, float),
+        Datatype.FP64: lambda x: isinstance(x, float),
+        Datatype.BYTES: lambda x: isinstance(x, (bytes, str)),
+    }
+
+    # look up and return the validation function for the expected_datatype if found, otherwise return None
+    is_datatype = type_checks.get(expected_datatype)
```
chang-annie, line 217: I: `type_checks` is dictionary where keys = the Datatype enums and values = lambda functions that check if given input is the correct Python type

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -126,3 +155,106 @@ class ServerMetadataResponse(BaseModel):
...
+
+
+def validate_shape(expected_shape: List[int], data: Union[List[Any], Any]) -> None:
+    """
+    Validates that the data (flat or nested) contains the correct number of elements based on the expected shape.
+    If nested, also ensures the structure is rectangular (no ragged lists) and matches the nested pattern.
+    """
+    # check that flattened data has the correct number of elements
+    flat = flatten(data)
+    expected_count = int(np.prod(expected_shape)) if expected_shape else 1
```
chang-annie, line 238: I: using `np.prod` to multiply the values in given `List[int] expected_shape` -> if `expected_shape = [2,2]`, then `expected_count = 4`

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -126,3 +155,106 @@ class ServerMetadataResponse(BaseModel):
...
+    # if data is nested and multidimensional, then check for rectangular structure & nested pattern
+    if (
+        expected_shape
+        and isinstance(data, list)
+        and len(expected_shape) > 1
+        and any(isinstance(elt, list) for elt in data)
+    ):
+        # np.array creates a true N-dimensional object array only if the data is fully rectangular
+        # otherwise, it returns an array of separate list objects.
+        data_array = np.array(data, dtype=object)
```
chang-annie, line 254: I: `np.array(data, dtype=object)` will try to convert the data into an `N-dimensional` object array
- if the data is truly rectangular, then it will return the N-dimensional array. 
  - for example, if `data = [[1,2],[3,4]]`
  ```
  [[1 2]
   [3 4]]
  ```
- otherwise, it returns an array of separate list objects
  - for example, if `data = [[1], [2,3]]`
  ```
  [list([1]) list([2, 3])]
  ```

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -66,10 +66,20 @@ def ModelInfer(
...
+            context.abort(
+                grpc.StatusCode.INTERNAL,
+                f"Failed to deserialize model inference request: {e}",
+            )
+            raise MappingException("Failed to deserialize model inference request!")
```
cwoods-cpointe: a: No need to raise because context.abort is already throwing exception

aissemble-open-inference-protocol-grpc/README.md
```diff
@@ -24,8 +24,12 @@ if __name__ == '__main__':
 The gRPC server will come up after a few seconds and will be OIP compliant. The proto specifications can be found in the [grpc_inference_service.proto](https://github.com/boozallen/aissemble-open-inference-protocol/blob/dev/aissemble-open-inference-protocol-grpc/proto/grpc_inference_service.proto) file.
 
 ### Implementing the Endpoints Handler
-By default, most of the gRPC endpoints will return a Method Not Implemented. You can implement these functions by creating a custom handler extending `DataplaneHandler`. Example:
+By default, most of the gRPC endpoints will return a Method Not Implemented. You can implement these functions by creating a custom handler extending `DataplaneHandler`.
 
+> [!NOTE]
+> All incoming `InferenceRequest` and outgoing `InferenceResponse` objects will be automatically validated against their declared tensor shapes and datatypes. Any discrepancy will raise an error and abort the call.
+
+Example:
```
cwoods-cpointe, line 32: a: I know its not your change but in the example we are setting - `datatype="FP32"`
but it should be - `Datatype.FP32` with the Datatype import

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -126,3 +155,106 @@ class ServerMetadataResponse(BaseModel):
...
+
+def get_actual_shape(data):
+    """
+    Returns a list of lengths at each nesting level. Assumes shape is rectangular.
+    """
+    shape = []
+
+    while isinstance(data, list):
+        shape.append(len(data))
+        data = data[0] if data else []
```
cwoods-cpointe: a: I think the `if data` will always be try because its an instance of list so not None

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -79,13 +90,31 @@ class InferenceResponse(BaseModel):
     parameters: Optional[Parameters] = None
     outputs: List[ResponseOutput]
 
+    def validate_oip(self) -> None:
+        if not self.outputs:
+            raise ValueError("InferenceResponse does not contain any outputs")
```
cwoods-cpointe: c/q: why is this a failure? if there was no output data it could mean the model has nothing to return but didnt fail

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -79,13 +90,31 @@ class InferenceResponse(BaseModel):
...
+        for idx, output_value in enumerate(self.outputs):
```
cwoods-cpointe: a: doesnt look like we are doing anything with the idx so no need to enumerate

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
chang-annie: good point! don't think it needs to raise an error. i'll update this to a log statement

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
chang-annie: yup good catch! leftovers from testing. removed `enumerate`

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
```diff
@@ -36,9 +36,6 @@
 from aissemble_open_inference_protocol_grpc.mappers.model_metadata_response_mapper import (
     ModelMetadataResponseMapper,
 )
-from aissemble_open_inference_protocol_grpc.mappers.utils import (
-    MappingException,
```
cwoods-cpointe, line 40: s: please remove class if no longer used

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
```diff
@@ -79,13 +93,34 @@ class InferenceResponse(BaseModel):
...
 class InferenceRequest(BaseModel):
     id: Optional[str] = None
     parameters: Optional[Parameters] = None
     inputs: List[RequestInput]
     outputs: Optional[List[RequestOutput]] = None
 
+    def validate_oip(self) -> None:
+        if not self.inputs:
+            raise ValueError("InferenceRequest does not contain any inputs")
+        for idx, input_value in enumerate(self.inputs):
```
cwoods-cpointe: a: not using idx so can probably remove the enumeration

cwoods-cpointe (approved):

jaebchoi (approved):

aissemble-open-inference-protocol-grpc/src/aissemble_open_inference_protocol_grpc/inference_servicer.py
chang-annie: already removed!

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/types/dataplane.py
chang-annie: removed!

Issue #76, opened 2025-08-11
Feature:Support rest of APIs for KServe integration with aiSSEMBLE OIP


jaebchoi:
## Description
As we have supported predict API for KServe integration with OIP, we need to supply additional API which is PreProcessing , Post Processing and load.

## DOD
- [x] Support Kserve's API under aiSSEMBLE OIP Handler
    - [x] Investigate whether this is logical distinction for pre/post processing. 
    - [x] Implement pre processing api
    - [x] Implement post processing api
    - [x] implement Load Api ( should be under -shared)
- [x] Add unit test if applicable
- [x] Start rough draft of README doc for Kserve
-------------------Stretch Goal------------
- [x] implement pre/post processing and load in the kserve Example.

### Prerequisite: 
To run this test, we need to build KServe Infrastructure App to test standing up KServe Correctly:
- Go to Branch `44-kserve-example`
- cd into `aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/infrastructure`
- Run following command to build dependent charts
```
helm dependency build ./cert-manager/
helm dependency build ./kserve-crds/
helm dependency build ./kserve/
helm dependency build ./nginx/
```
NOTE: make sure you have following repo added in your helm cli
```
stable    	https://charts.helm.sh/stable
bitnami    	https://charts.bitnami.com/bitnami
apache-airflow	https://airflow.apache.org/
elastic    	https://helm.elastic.co/
jenkins    	https://charts.jenkins.io/
spark-operator	https://kubeflow.github.io/spark-operator/
localstack  	https://localstack.github.io/helm-charts
hashicorp   	https://helm.releases.hashicorp.com/
argoCD    	https://argoproj.github.io/argo-helm
ingress-nginx 	https://kubernetes.github.io/ingress-nginx
trino     	https://trinodb.github.io/charts
istio     	https://istio-release.storage.googleapis.com/charts
jetstack   	https://charts.jetstack.io/

```
- Run following command to install KServe infrastructures
    
```
helm install kserve-crds ./kserve-crds/
helm install ingress-nginx ./nginx/ --namespace ingress-nginx --create-namespace
helm install cert-manager ./cert-manager/ --create-namespace --namespace cert-manager
helm install kserve ./kserve/
```
### Test Steps
Once all installation is complete you can proceed with testing steps
- Switch to dev branch.
[...]

PR #79, opened 2025-08-18
[#76] support rest of kserve apis

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/76


aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
```diff
@@ -44,6 +44,10 @@ def predict(
         )
         return infer_response
 
+    def load(self):
+        self.ready = True
```
aaron-gary, line 50: Q: Does this need a comment to let the user know they should handle the model loading before setting ready to true?

aaron-gary (approved):

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
jaebchoi: I actually have them in the README, but I can def add them here too!

PR #87, opened 2025-08-27
[#82] address issues in release.yaml

### LINKED ISSUE #82: Feature: Release aiSSEMBLE OIP FastAPI / gRPC/KServe/Shared
## Description
As we have finished most of work in FASTAP, gRPC, KServe and Shared Module of aiSSEMBLE OIP, we would like to release them for consumer to use.

## DOD
- [x] add github action to release the aissemble-open-inference-protocol repository.
    - [x] look into how to get credentials for pypi 
- [x] Release [aissemble-open-inference-protocol](https://github.com/boozallen/aissemble-open-inference-protocol)
- [x] Add Release Note ( follow habushu or aissemble release note template)
- [x] Verify modules (FastAPI/gRPC/KServe/Shared) are uploaded to pypi.
    - [x] Make sure modules in the pypi is under boozallen org. 
    - [x] Make sure pypi metadata is correct
- [x] Make sure mentioned maven packages are also published and investigate where it shall publish( we think it's mavn central) 

## Test Strategy/Script
- [x] Verify [aissemble-open-inference-protocol](https://github.com/boozallen/aissemble-open-inference-protocol) is released to 1.0.1
- [x] Verify 1.0.1 have release notes
- [x] Verify you can see [aissemble-open-inference-protocol-fastapi](https://github.com/boozallen/aissemble-open-inference-protocol/tree/dev/aissemble-open-inference-protocol-fastapi), [aissemble-open-inference-protocol-grpc](https://github.com/boozallen/aissemble-open-inference-protocol/tree/dev/aissemble-open-inference-protocol-grpc) [aissemble-open-inference-protocol-kserve](https://github.com/boozallen/aissemble-open-inference-protocol/tree/dev/aissemble-open-inference-protocol-kserve) and [aissemble-open-inference-protocol-shared](https://github.com/boozallen/aissemble-open-inference-protocol/tree/dev/aissemble-open-inference-protocol-shared)  module are available through pypi. see [Here](https://pypi.org/search/?q=aissemble-open-inference)
- [x] Verify sonatype maven centrals have all aissemble oip modules see [here](https://central.sonatype.com/search?q=aissemble-open-inference) 
- [x] Verify [Build](https://github.com/boozallen/aissemble-open-inference-protocol/actions/workflows/build.yaml) job is successful

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/82

### DIFF
--- a/.github/actions/install_dependencies/action.yaml
+++ b/.github/actions/install_dependencies/action.yaml
@@ -1,0 +1,58 @@
+name: 'Install Dependencies'
+description: 'Install the dependencies needed to build aissemble-open-inference-protocol'
+inputs:
+  sonatype-central-user:
+    required: true
+  sonatype-central-key:
+    required: true
+  github-actor:
+    required: true
+  github-token:
+    required: true
+  pypi-username:
+    required: true
+  pypi-token:
+    required: true
+  package-index:
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11.4'
+    - name: Install Poetry
+      uses: snok/install-poetry@v1
+    - name: Set up JDK 17
+      uses: actions/setup-java@v4
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Create settings.xml
+      id: create-settings-xml
+      shell: bash
+      run: |
+        cat > $HOME/.m2/settings.xml << EOF
+        <settings>
+          <servers>
+            <server>
+              <id>central</id>
+              <username>${{ inputs.sonatype-central-user }}</username>
+              <password>${{ inputs.sonatype-central-key}}</password>
+            </server>
+            <server>
+              <id>ghcr.io</id>
+              <username>${{ inputs.github-actor }}</username>
+              <password>${{ inputs.github-token }}</password>
+            </server>
+            <server>
+              <id>${{ inputs.package-index }}</id>
+              <username>${{ inputs.pypi-username }}</username>
+              <password>${{ inputs.pypi-token }}</password>
+            </server>
+          </servers>
+        </settings>
+        EOF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,10 +12,9 @@ on:
   workflow_dispatch:
     inputs:
       releaseBranch:
-        description: "Release Branch"
-        required: true
+        description: "Release Branch (default: release-{releaseVersion})"
+        required: false
         type: string
-        default: "release-RELEASE_VERSION"
       releaseVersion:
         description: "Release Version (X.Y.Z)"
         required: true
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -86,7 +86,6 @@ jobs:
         run: |
           git config --local user.email "aissemble@bah.com"
           git config --local user.name "Github Actions"
-          git checkout -b ${{ input.releaseBranch }}
           mvn -B release:clean release:prepare -Prelease-ci -DreleaseVersion=${{ env.RELEASE_VERSION }} -DdevelopmentVersion=${{ env.NEXT_DEVELOPMENT_VERSION }} -Dmaven.build.cache.enabled=false
```
jaebchoi: I: it should be `inputs.releaseBranch` not `input` but we just need to clean this line up as we did those ref in the checkout@v4 step

epwilkins: It looks like the profiles passed to Maven weren't updated as well

jaebchoi: ah def missed that. good catch!

.github/actions/install_dependencies/action.yaml
```diff
@@ -0,0 +1,56 @@
+name: 'Install Dependencies'
+description: 'Install the dependencies needed to build aissemble-open-inference-protocol'
```
epwilkins, line 2: S: Use this in build.yaml as well

jaebchoi: Settings.xml is different in build vs release i.e. build.yaml doesn't have sonatype entries.

So in this case, Ill pull out settings.xml from action.yaml unless there's way to do different versions of setting.xml in composite run

jaebchoi: maybe we can do that using` [if input is release or build] && {do sth} || {dosth} ` but seems bit messy so Ill pull out settings.xml one!

epwilkins: no I think it's fine to just let the build have the sonatype server setup

epwilkins: But if we aren't using this composite action in both build and release, there isn't much use in pulling it out.

.github/workflows/release.yaml
```diff
@@ -12,8 +12,8 @@ on:
   workflow_dispatch:
     inputs:
       releaseBranch:
-        description: "Release Branch"
-        required: true
+        description: "Release Branch (default: release-{releaseVersion})"
+        required: false
         type: string
         default: "release-RELEASE_VERSION"
```
epwilkins: A: remove

.github/actions/install_dependencies/action.yaml
jaebchoi: I see, okay Ill keep the settings.xml in composite

.github/actions/install_dependencies/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+inputs:
+  sonatype-central-user:
+    required: true
```
aaron-gary, line 5: Q: Are these inputs a temporary fix?  i.e. shouldn't they be pulled from the secrets?

aaron-gary: OK, after seeing the way it's being used, I assume the main use is to call it from another script.

aaron-gary (approved):

.github/actions/install_dependencies/action.yaml
jaebchoi: Yup! we pass in secrets in build.yaml or release.yaml so it's encrypted

jaebchoi:
- Updated to add setting.xml back to. composite run
- Verified [build](https://github.com/boozallen/aissemble-open-inference-protocol/actions/runs/17275090497/job/49029707109) run passes.

epwilkins (approved):

PR #89, opened 2025-08-28
[#82] Post Release version bump and fix release job

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/82

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,24 +7,17 @@
     <parent>
         <groupId>com.boozallen.aissemble</groupId>
         <artifactId>aissemble-parent</artifactId>
-        <version>6</version>
+        <version>7</version>
         <relativePath />
     </parent>
 
     <artifactId>aissemble-open-inference-protocol</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.1.0-SNAPSHOT</version>
 
     <name>aiSSEMBLE::Open Inference Protocol</name>
     <description>Houses all of aiSSEMBLE's implementations of the Open Inference Protocol (OIP)</description>
     <packaging>pom</packaging>
     <url>https://github.com/boozallen/aissemble-open-inference-protocol</url>
-
-    <distributionManagement>
-        <snapshotRepository>
-            <id>ghcr.io</id>
-            <url>https://maven.pkg.github.com/boozallen/aissemble-open-inference-protocol</url>
-        </snapshotRepository>
-    </distributionManagement>
 
     <scm>
         <connection>scm:git:https://github.com/boozallen/aissemble-open-inference-protocol.git</connection>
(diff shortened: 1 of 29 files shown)


.github/actions/install_dependencies/action.yaml
```diff
@@ -43,11 +43,6 @@ runs:
               <username>${{ inputs.sonatype-central-user }}</username>
               <password>${{ inputs.sonatype-central-key}}</password>
             </server>
-            <server>
-              <id>ghcr.io</id>
-              <username>${{ inputs.github-actor }}</username>
-              <password>${{ inputs.github-token }}</password>
-            </server>
```
jaebchoi: I: Don't need to publish ghcr no more per discussion

pom.xml
```diff
@@ -7,25 +7,18 @@
     <parent>
         <groupId>com.boozallen.aissemble</groupId>
         <artifactId>aissemble-parent</artifactId>
-        <version>6</version>
+        <version>7</version>
```
jaebchoi, line 10: I: version 7 includes changing ossrh into sonatype so you need 7 to publish central sonatype.

pom.xml
```diff
@@ -55,53 +48,30 @@
...
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>license-maven-plugin</artifactId>
-            </plugin>
-        </plugins>
-    </build>
-
     <profiles>
+        <profile>
+            <id>default-build</id>
```
jaebchoi, line 59: I: had to create default build profile and add all modules as modules without profile will always build all modules regardless of profiles,  ( we want release profile to exclude examples)

pom.xml
```diff
@@ -7,25 +7,18 @@
...
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.1.0-SNAPSHOT</version>
 
     <name>aiSSEMBLE::Open Inference Protocol</name>
     <description>Houses all of aiSSEMBLE's implementations of the Open Inference Protocol (OIP)</description>
     <packaging>pom</packaging>
     <url>https://github.com/boozallen/aissemble-open-inference-protocol</url>
 
-    <distributionManagement>
-        <snapshotRepository>
```
aaron-gary, line 23: Q: What happens when we remove this snapshotRepository?  Does it default to maven central?

aaron-gary (approved):

pom.xml
cwoods-cpointe: we should be able to achieve this by using the maven-release-plugin or Habushu. Having to keep 2 lists of modules in sync can cause overhead

cwoods-cpointe:
Lets create a draft release notes to prep for future releases and keep release docs aligned

pom.xml
jaebchoi: I think it kicks off habushu artifact deploy on release, but it does fail at Build.yaml where it executes mvn default deploy so added new profile for build to add this.

pom.xml
jaebchoi: Ill investigate and do follow up with this!

PR #93, opened 2025-09-02
[#82] fix missed version bump

jaebchoi:
we have failed build in mvn clean install due to not doing version bump on one module.
This is fixing that as well as refactoring profile for cleaner code.

### DIFF
--- a/aissemble-open-inference-protocol-examples/aissemble-oip-grpc/pom.xml
+++ b/aissemble-open-inference-protocol-examples/aissemble-oip-grpc/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.boozallen.aissemble</groupId>
         <artifactId>aissemble-open-inference-protocol-examples</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <version>1.1.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>aissemble-oip-grpc</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -54,26 +54,20 @@
         <version.python>3.11.4</version.python>
     </properties>
 
+    <modules>
+        <module>aissemble-open-inference-protocol-fastapi</module>
+        <module>aissemble-open-inference-protocol-grpc</module>
+        <module>aissemble-open-inference-protocol-shared</module>
+        <module>aissemble-open-inference-protocol-examples</module>
+        <module>aissemble-open-inference-protocol-kserve</module>
+    </modules>
+
     <profiles>
         <profile>
             <id>default-build</id>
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -71,9 +71,6 @@
                 <module>aissemble-open-inference-protocol-examples</module>
                 <module>aissemble-open-inference-protocol-kserve</module>
             </modules>
-        </profile>
-        <profile>
-            <id>build</id>
```
epwilkins, line 76: Q: If this is the default build, why wrap it in a profile?

aissemble-open-inference-protocol-examples/aissemble-oip-grpc/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.boozallen.aissemble</groupId>
         <artifactId>aissemble-open-inference-protocol-examples</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <version>1.1.0-SNAPSHOT</version>
```
epwilkins, line 10: Q: was this missed because the examples aren't part of the release modules? If so, might make sense to include them, but add a `release` profile definition here that excludes its children, or otherwise disables publishing.

pom.xml
jaebchoi: if I don't wrap the profile, it would always kick off property that are in the outside of profile, so any property inside of profile doesn't get overridden from my testing.

aissemble-open-inference-protocol-examples/aissemble-oip-grpc/pom.xml
jaebchoi: yeah, so I had to manually upversion them, `release:prepare` bumps up the version for you. Ill reach out for detail! not sure I understood about exclude children/disable publish

pom.xml
epwilkins: property or tag?  I'd think property values should override just fine as we do that all the time. Tags can be a bit trickier, but you should be able to include `combine.self="override"` as an attribute on the tag (see [docs](https://maven.apache.org/pom.html#Advanced_configuration_inheritance.3A_combine.children_and_combine.self)).

My primary concern is that it looks like a developer would need to always add this profile to their build, which is a pattern that often causes confusion and issues in my experience.  Best if the actual default configuration works for local building, and profiles are layered to customize for things like CI.

jaebchoi:
TLDR; Per discussion, we decided to include all modules in root pom.
In the -example pom we configured to skip publishing to pypi and central (Not fully tested, next person who is doing release shall test it).

epwilkins (approved):

PR #99, opened 2025-09-08
[#90] Allow custom handlers to override model load

### LINKED ISSUE #90: Feature: Model loading should be in the handler to allow custom loading logic
## Description
In #76 we added `load` into the `AissembleOIPService` base class. This forced the child fastapi, grpc, kserve impls to have this load function but they are not qualified to know what the load impl is. This load function needs to be implemented by the user.

**Technical discussion:**
Users can call load when its best for them. They have control over predict, model_ready, etc so they can call it whenever. We can assume the load is eager and have the default model_ready return true.

Kserve impl - Us and the user are not in control of when load is called so no need for us to do anything fancy.

## DOD
- [x] Allow users to implement model loading
  - [x] The abstract load function is in the base handler class
  - [x] There is no default behavior
- [x] The model_ready default call is updated to assume the model was loaded eagerly and return true
- [x] Documentation and examples of this behavior is available
  - [x] Fastapi
  - [x] gRPC
  - [x] Kserve - Note the Kserve example is outdated so hard to test. Will be addressed on follow on

## Test Strategy/Script
1. Build the project to get the examples up to date - `mvnd clean install`
2. Navigate to the fastapi examples - `cd aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference` 
3. Verify the handler in the `main.py` file is overriding the `model_load` function
4. Start the fastAPI aissemble-oip-fastapi-inference app - `poetry run run_server`
5. Call the `Model Ready` endpoint as described in the readme
6. Verify the response is true
7. Call each of the other endpoints as described in the readme
8. Verify the responses are successful
9. Navigate to the gRPC inference example - `cd aissemble-open-inference-protocol-examples/aissemble-oip-grpc/aissemble-oip-grpc-inference`
10. Verify the `oip_handler.py` file is overriding the `model_load` function
11. Start the grpc server with `poetry run run_server`
12. Verify the logs show `OIPHandler: Model has been loaded.`
13. Run each of the grpcurl commands in the README
14. Verify each are successful

See #76

cwoods-cpointe:
Model loading should be implementable by downstream handlers. The calling of model loading should be up to the downstream so we will not be calling load.

Updated the examples to reflect this change. Also changed the Kserve example to no longer extend the base aissemble oip kserve class

[#90] simplified and corrected fastapi example


aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference/src/aissemble_oip_fastapi_inference/main.py
```diff
@@ -36,16 +38,17 @@ class Handler(DataplaneHandler):
 
     def __init__(self):
         super().__init__()
+        self.model = None
+        self.mode_ready = self.model_load("convert_celsius_to_fahrenheit")
```
cwoods-cpointe: i: The way we are injecting the handler with fastapi requires the load to be ran on each call. This is not ideal and will be addressed in a follow on

epwilkins: A:
```suggestion
        self.model_ready = self.model_load("convert_celsius_to_fahrenheit")
```

epwilkins: What's the high-level explanation for why this is?

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
```diff
@@ -109,80 +103,16 @@ def model_ready(
...
-        max_asyncio_workers=model.config.kserve_max_asyncio_workers,
-        enable_grpc=model.config.kserve_enable_grpc,
-        enable_docs_url=model.config.kserve_enable_docs_url,
-        enable_latency_logging=model.config.kserve_enable_latency_logging,
-        access_log_format=model.config.kserve_access_log_format,
-        grace_period=model.config.kserve_grace_period,
-    ).start([model])
+    model_name = "convert_celsius_to_fahrenheit"
+    oip_kserve = AissembleOIPKServe(name=model_name, handler=CustomFastAPIHandler())
+    oip_kserve.model_load(model_name=model_name)
```
epwilkins: Q: Why is it necessary to call this explicitly?

aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference/src/aissemble_oip_fastapi_inference/main.py
cwoods-cpointe: When a request is made, fastapi will instantiate a new handler. So if users call handler.load outside of that call, the results will be ignored

epwilkins (approved):

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
cwoods-cpointe: hmm good point. I was originally thinking Kserves model.load wouldnt work but let me test it again. Ideally Kserves call to its model.load would route to the users handler

cwoods-cpointe: Corrected the behavior so our class extending kserves model will call the handler on kserve load. Even though kserve will call the load if the model has not been registered, I will still leave this load here as per the kserve [instructions](https://kserve.github.io/website/docs/model-serving/predictive-inference/frameworks/custom-predictor).

Issue #102, opened 2025-09-08
Feature: Correct KServe example 


csun-cpointe:
## Description
Follow on to the #90 , when starting the kserve example service, the service will fail at repeatedly calling the KServe model_server.start() function. We need to figure out why the repeatedly start() calls and fix the kserve example.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Fix the Kserve example
- [x] Code update if necessary

## Test Strategy/Script
**Set up**
1. Pull down changes and build the project - `mvnd clean install`
2. Download the [kserve-example.zip](https://github.com/user-attachments/files/22174575/kserve-example.zip) file attached to the ticket and deploy KServe dependencies:
    1. Change to the `kserve-example/deploy` directory and run:
      ```
      helm dependency build ./infrastructure/cert-manager/
      helm dependency build ./infrastructure/kserve-crds/
      helm dependency build ./infrastructure/kserve/
      helm dependency build ./infrastructure/nginx/
      ``` 
    2. Install the services:
      ```
      helm install kserve-crds ./infrastructure/kserve-crds/
      helm install ingress-nginx ./infrastructure/nginx/ --namespace ingress-nginx --create-namespace
      helm install cert-manager ./infrastructure/cert-manager/ --create-namespace --namespace cert-manager
      helm install kserve ./infrastructure/kserve/
      ```
3. Create the namespace: `kubectl create namespace kserve-test`

**Test http**

4. Replace the contents of the `aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-containerization/aissemble-oip-kserve-containerization-deploy/src/main/resources/templates/inference-service.yaml` with the following:    
    ```
    apiVersion: serving.kserve.io/v1beta1
    kind: InferenceService
    metadata:
      name: kserve-model
    spec:
      predictor:
        containers:
          - name: "aissemble-oip-kserve-containerization-docker"
            image: aissemble-oip-kserve-containerization-docker:1.1.0-SNAPSHOT
            env:
              - name: KSERVE_HTTP_PORT
                value: "9090"
            ports:
              - containerPort: 9090
                protocol: TCP
            readinessProbe:
              httpGet:
                path: /v2/health/ready
                port: 9090
              initialDelaySeconds: 10
              periodSeconds: 5
              timeoutSeconds: 3
              failureThreshold: 3
    ```
[...]

PR #104, opened 2025-09-10
[#92] add custom handler for kserve to ease switching between protocols

### LINKED ISSUE #92: Feature: Determine how to use handler functions in KServe apps when switching between implementations
## Description
Typically, KServe will call out to its own `Model` class to run predict, load, etc. Currently for aiSSEMBLE KServe `predict` method, we intercept the call and send it our handler.

We need to determine how to implement the rest of the handler functions in KServe apps to reduce cognitive friction when users switch between different implementations (FastAPI, gRPC, KServe, etc)

## DOD
 - [x] AissembleOIPKServe for ModelServer uses our customKServeDataplaneHandler
     - [x] DataplaneHandler should extend KServe's Dataplane Class
 - [x] Make sure All OIP Endpoints using KServe uses our customKServeDataplaneHandler logic
     - [x] No Behavior changes upon REST /gRPC on KServe's environment
===========Possible split Point=========================
-  [ ] think of passing in share dataplane handler into KserveDataPlaneHandler so that we can just map response into what KServe Expects.
- [ ] Update Kserve's README with proposed architecture.
- [ ] Think about How we can abstract out Shared vs KServe Dataplane Handler. 

### KServe Installation Setup
- Pull latest [feature/kserve-custom-handler](https://github.com/boozallen/aissemble-open-inference-protocol/tree/feature/kserve-custom-handler) branch build the project - `mvnd clean install`
- Download [kserve-example.zip](https://github.com/user-attachments/files/22238326/kserve-example.zip)
- unzip and cd into `kserve-example/deploy` directory and run:
```
helm dependency build ./infrastructure/cert-manager/
helm dependency build ./infrastructure/kserve-crds/
helm dependency build ./infrastructure/kserve/
helm dependency build ./infrastructure/nginx/
``` 
- Install the services:
```
helm install kserve-crds ./infrastructure/kserve-crds/
helm install ingress-nginx ./infrastructure/nginx/ --namespace ingress-nginx --create-namespace
helm install cert-manager ./infrastructure/cert-manager/ --create-namespace --namespace cert-manager
helm install kserve ./infrastructure/kserve/
```
- Create the namespace: `kubectl create namespace kserve-test`

### Test Deploying KServe
- cd into kserve deployment charts 
` cd aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-containerization/aissemble-oip-kserve-containerization-deploy/src/main/resources/templates`
- Apply Inference service helm chart to kserve-test namespace
` kubectl apply -n kserve-test -f ./inference-service.yaml`
- check status of pods until predictor pod is ready
`kubectl get pods -n kserve-test`
- On the Rancher, go to port fowarding and check if kserve-model-predictor is forwarding to 8080 if not set it to 8080

Image

### Test KServe OIP Endpoints
- Run Curl Command to send inference Request
```

curl --request POST \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2/models/convert_celsius_to_fahrenheit/infer \
--data '{
"id" : "1",
"inputs" : [
{
"name" : "sample",
"shape" : [1],
"datatype"  : "FP32",
"parameters": {
"content_type": "str"
},
"data" : [[ 0]]
}
]
}'
```
- Verify Response something like this:
```
{"model_name":"kserve-model","model_version":null,"id":"1","parameters":null,"outputs":[{"name":"kserve-model","shape":[1],"datatype":"FP32","parameters":null,"data":[31.86412239074707]}]}% 
```

- Run Curl commands to send Following Curl Commands and Verify Successful:

**Model Ready**
```
curl --request GET \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2/models/convert_celsius_to_fahrenheit/ready
```
**Model Metadata**
```
curl --request GET \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2/models/convert_celsius_to_fahrenheit
```
**Server Ready**
```
curl --request GET \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2/health/ready
```
**Server Live**
```
curl --request GET \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2/health/live
```
**Server Metadata**
```
curl --request GET \
-H "Content-Type: application/json" \
--url http://localhost:8080/v2
```

Add any other context, links, or screenshots about the feature request here.

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/92

### DIFF
--- a/aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
+++ b/aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
@@ -8,23 +8,22 @@
 # #L%
 ###
 import numpy as np
-from typing import Optional
+from typing import Optional, Union, Dict, Tuple
 
 from tensorflow.keras.models import load_model
+
+from kserve import InferRequest, InferResponse, InferOutput, ModelServer
 
 from aissemble_open_inference_protocol_kserve.aissemble_oip_kserve import (
     AissembleOIPKServe,
 )
 
-from aissemble_open_inference_protocol_shared.handlers.dataplane import DataplaneHandler
 from aissemble_open_inference_protocol_shared.types.dataplane import (
-    InferenceRequest,
-    InferenceResponse,
-    ModelMetadataResponse,
-    ModelReadyResponse,
-    MetadataTensor,
-    ResponseOutput,
     Datatype,
+)
+
+from aissemble_open_inference_protocol_kserve.kserve_dataplane import (
+    KServeDataplaneHandler,
 )
 
 """"
@@ -34,76 +33,76 @@ to the AissembleOIPKServe constructor along with the model name. In this example
 """
 
 
-class CustomFastAPIHandler(DataplaneHandler):
+class CustomKServeDataplaneHandler(KServeDataplaneHandler):
     def __init__(self):
         super().__init__()
         self.ready = False
         self.model = None
 
-    def infer(
+    async def infer(
         self,
-        payload: InferenceRequest,
         model_name: str,
-        model_version: Optional[str] = None,
-    ) -> InferenceResponse:
+        request: Union[Dict, InferRequest],
+        headers: Optional[Dict[str, str]] = None,
+    ) -> Tuple[Union[Dict, InferResponse], Dict[str, str]]:
+        keras_model = load_model("model/" + model_name + ".keras")
         # Model will take input data from the payload and make prediction to convert Celsius to Fahrenheit.
-        output = self.model.predict(np.array(payload.inputs[0].data))
+        output = keras_model.predict(np.array(request.inputs[0].data))
         # Need to convert to list so that we are align with output format.
         output_list = output.tolist()
-
-        return InferenceResponse(
+        response_headers = {}
+        return InferResponse(
+            response_id=request.id,
             model_name=model_name,
-            model_version=model_version,
-            id=payload.id,
-            outputs=[
-                ResponseOutput(
+            infer_outputs=[
+                InferOutput(
                     name=model_name,
-                    shape=payload.inputs[0].shape,
+                    shape=request.inputs[0].shape,
                     datatype=Datatype.FP32,
                     data=output_list,
                 )
             ],
-        )
+        ), response_headers
 
-    def model_metadata(
+    async def model_metadata(
         self,
         model_name: str,
         model_version: Optional[str] = None,
-    ) -> ModelMetadataResponse:
+    ) -> Dict:
+        load_model("model/" + model_name + ".keras")
         input_tensors = []
         for input in self.model.inputs:
             datatype = None
             if input.dtype == "float32":
-                datatype = Datatype.FP32
-            inputmtensor = MetadataTensor(
-                name="input", datatype=datatype, shape=[input.shape[1]]
-            )
-            input_tensors.append(inputmtensor)
-
+                datatype = "FP32"
+            input_dict = {
+                "name": "input",
+                "datatype": datatype,
+                "shape": [input.shape[1]],
+            }
+            input_tensors.append(input_dict)
         output_tensors = []
         for output in self.model.outputs:
             datatype = None
             if output.dtype == "float32":
-                datatype = Datatype.FP32
-            outputmtensor = MetadataTensor(
-                name="output", datatype=datatype, shape=[output.shape[1]]
-            )
-            output_tensors.append(outputmtensor)
+                datatype = "FP32"
+            output_dict = {
+                "name": "output",
+                "datatype": datatype,
+                "shape": [output.shape[1]],
+            }
+            output_tensors.append(output_dict)
+        return {
+            "name": model_name,
+            "platform": "python",
+            "inputs": input_tensors,
+            "outputs": output_tensors,
+        }
 
-        return ModelMetadataResponse(
-            name=model_name,
-            versions=[model_version] if model_version else None,
-            platform="python",
-            inputs=input_tensors,
-            outputs=output_tensors,
-        )
-
-    def model_ready(
-        self,
-        model_name: str,
-        model_version: Optional[str] = None,
-    ) -> ModelReadyResponse:
-        return ModelReadyResponse(name=model_name, ready=self.ready)
+    async def model_ready(
+        self, model_name: str, disable_predictor_health_check: bool = False
+    ) -> bool:
+        return True
 
     def model_load(self, model_name) -> bool:
         self.model = load_model("model/" + model_name + ".keras")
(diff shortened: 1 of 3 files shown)


jaebchoi:
I: This is merging into Feature branch as I need follow up ticket to  passing in share dataplane handler into KserveDataPlaneHandler.

This Ticket's DPC is to make sure we can override our custom KServeDataPlaneHandler from Kserve's model server and Verify OIP Endpoint is still able to function with our custom KServeDataPlaneHandler.

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/kserve_dataplane.py
```diff
@@ -0,0 +1,112 @@
...
+# This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+# #L%
+###
+from typing import Dict, Union, Optional, Tuple
+
+from kserve import InferRequest, InferResponse, ModelRepository
+from kserve.protocol.dataplane import DataPlane
+
+
+class KServeDataplaneHandler(DataPlane):
```
jaebchoi, line 16: I: for now, those implementations are just default values. But on follow up ticket it should call our DataplaneHandler so that client doesn't have to implement both KserveDataplaneHandler and DataplaneHandler, rather they just focus on DataplaneHandler

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
```diff
@@ -8,102 +8,107 @@
...
 ###
 import numpy as np
-from typing import Optional
+from typing import Optional, Union, Dict, Tuple
 
 from tensorflow.keras.models import load_model
 
+from kserve import InferRequest, InferResponse, InferOutput, ModelServer
+
+from krausening.logging import LogManager
```
cwoods-cpointe: a: unused import

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/kserve_dataplane.py
cwoods-cpointe: i: Further after that we will investigate removing the need for users to create a DataplaneHandler by creating our own default handlers. Then users will only need to create the model functions like infer and load

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
```diff
@@ -8,102 +8,107 @@
...
+        ), response_headers
 
-    def model_metadata(
+    async def model_metadata(
         self,
         model_name: str,
         model_version: Optional[str] = None,
-    ) -> ModelMetadataResponse:
+    ) -> Dict:
+        model = load_model("model/" + model_name + ".keras")
```
cwoods-cpointe: a: model is unused

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
```diff
@@ -8,102 +8,107 @@
...
-        model_name: str,
-        model_version: Optional[str] = None,
-    ) -> ModelReadyResponse:
-        return ModelReadyResponse(name=model_name, ready=self.ready)
+    async def model_ready(
+        self, model_name: str, disable_predictor_health_check: bool = False
+    ) -> bool:
+        try:
+            load_model("model/" + model_name + ".keras")
+            return True
```
cwoods-cpointe: a: can just return self.ready

aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-inference/src/aissemble_oip_kserve_inference/main.py
```diff
@@ -113,6 +118,20 @@ def model_load(self, model_name) -> bool:
 
 if __name__ == "__main__":
     model_name = "convert_celsius_to_fahrenheit"
-    oip_kserve = AissembleOIPKServe(name=model_name, handler=CustomFastAPIHandler())
+    oip_kserve = AissembleOIPKServe(name=model_name, handler =CustomKServeDataplaneHandler(),
+    )
     oip_kserve.load()
-    oip_kserve.start()
+    model_server = ModelServer(
```
cwoods-cpointe, line 120: i: oip_kserve.start is what we ultimately want to use but it has a bug thats being addressed in #102

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
```diff
@@ -7,47 +7,33 @@
...
 class AissembleOIPKServe(Model, AissembleOIPService):
-    def __init__(self, name: str, handler: DataplaneHandler = DefaultHandler()):
+    def __init__(
+        self, name: str, handler: KServeDataplaneHandler = KServeDataplaneHandler()
+    ):
         Model.__init__(self, name)
         AissembleOIPService.__init__(self, handler=handler, adapter=None)
         self.model = None
 
-    def predict(
```
cwoods-cpointe, line 28: q/c: why were load and predict changed? Id assume this class wouldnt change outside of the handler

cwoods-cpointe: Ah is the KServeDataplaneHandler not calling this predict but just its own infer? Would still need load

jaebchoi: - For predict: as we discussed before, here were the flow 
**Before we create Kserve_dataplane (Kserve's default behavior):**
Dataplane infer() will call model with InferRequest -> model will call predict/preprocess/postprocess and do its logic -> then get inferResponse
**When we create our custom Kserve Dataplane:**
Our KserveDataplaneHandler takes care of infer() do our logic and return inferResponse without going to predict 

So I was thinking if we intercept infer() and do all our logic there, predict will never get called unless we explictly call model or use KServe's default behavior 

- For Load: KServeDataplaneHandler class doesn't have model_load() method, I think we can keep this impl by adding model_load function in the kserve_dataplane.py

jaebchoi: Updated the load logic!

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
```diff
@@ -7,47 +7,30 @@
...
-        )
-        return infer_response
-
     def load(self) -> bool:
         return self.handler.model_load(self.name)
 
-    def start(self):
-        ModelServer(
+    async def start(self):
+        self.load()
```
cwoods-cpointe: s: we are leaving the loading calls up to the user (at least for now) as they know best when/where to load their models.

cwoods-cpointe (approved):

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/kserve_dataplane.py
epwilkins: It seems to me that we shouldn't have a special class that they have to extend just for KServe. We should create something like a `DataPlaneHandlerAdapter` that exends KServe's `DataPlane` wraps the generic `aissemble-oip` DPH and correctly routes logic where appropriate.  Then, we could wrap the user-provided DPH in the Adapter during the construction of the Service, making the process totally transparent to the user.

epwilkins: I'm also confused why our KServe service is still extending `Model`. That seems incorrect with this new approach.

jaebchoi: > It seems to me that we shouldn't have a special class that they have to extend just for KServe. We should create something like a `DataPlaneHandlerAdapter` that exends KServe's `DataPlane` wraps the generic `aissemble-oip` DPH and correctly routes logic where appropriate. Then, we could wrap the user-provided DPH in the Adapter during the construction of the Service, making the process totally transparent to the user.

Yup that is what I am trying to do on the follow up ticket. This ticket I should say is experimenting that overriding KServe's DataPlane is actually doable so that we can start wrapping with our aissemble DPH.

jaebchoi: > I'm also confused why our KServe service is still extending `Model`. That seems incorrect with this new approach.

Main reason with Extending Model is due to load() class. We need to make sure Model's ready attribute should set to True when we loading otherwise it will fail at this [point](https://github.com/kserve/kserve/blob/3efa61b72a2860e97ea08515fb324874f0b60fe3/python/kserve/kserve/model_server.py#L408)  Happy to hear different approach if you can think of other!

Issue #112, opened 2025-09-15
TASK: Fix broken build


d-ryan-ashcraft:
## Description
The build is currently failing:
```
Run mvn clean deploy -B -U -Pdefault-build,ci
  mvn clean deploy -B -U -Pdefault-build,ci
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.4/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.4/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.4/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.4/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.4/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.4/x64/lib
    VENV: .venv/bin/activate
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.16-8/x64
    JAVA_HOME_17_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.16-8/x64
    GITHUB_TOKEN: ***
[INFO] Loading cache configuration from /home/runner/work/aissemble-open-inference-protocol/aissemble-open-inference-protocol/.mvn/maven-build-cache-config.xml
[INFO] Using SHA-256 hash algorithm for cache
[INFO] Scanning for projects...
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/technologybrewery/fabric8/docker-maven-plugin/maven-metadata.xml
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/technologybrewery/fabric8/docker-maven-plugin/maven-metadata.xml (379 B at 630 B/s)
Warning:  
Warning:  Some problems were encountered while building the effective model for com.boozallen.aissemble:aissemble-oip-kserve-containerization-docker:docker-build:1.1.0-SNAPSHOT
Warning:  'build.plugins.plugin.version' for org.technologybrewery.fabric8:docker-maven-plugin is missing. @ line 40, column 21
Warning:  
Warning:  It is highly recommended to fix these problems because they threaten the stability of your build.
Warning:  
Warning:  For this reason, future Maven versions might no longer support building such malformed projects.
Warning:  
[...]

PR #113, opened 2025-09-16
[#112] prevent using poetry 2.2.0

jaebchoi:
https://github.com/boozallen/aissemble-open-inference-protocol/issues/112


d-ryan-ashcraft:
S (Substantive): won't this PR also need the license file updates in the `pyproject.toml` files to pass?

jaebchoi:
> S (Substantive): won't this PR also need the license file updates in the `pyproject.toml` files to pass?

That seems to be enforced starting 2.2.0, so we seem to pass the build at this moment.
But with that said,  we can add this now so that we can have smooth sailing when upgrading to 2.2.0 later

d-ryan-ashcraft (approved):

Issue #100, opened 2025-09-08
Feature: KServe config resolution should support args in inference service yaml


colinpalmer-pro:
[kserve-example.zip](https://github.com/user-attachments/files/22174575/kserve-example.zip)

## Description
In #85, we enabled KServe server configs to be set via krausening properties and environment variables to align with the broader OIP configuration approach utilized with FastAPI and gRPC. However, this change no longer honors KServe's default behavior of handling configs via `args` in the `InferenceService` definition. 

This ticket adds support for args-based configuration with the following precedence: args> env variables > krausening properties. This ensures consistency with KServe’s defaults while maintaining alignment across OIP offerings.

## DOD
- [x] KServe server configs can be set via args in the InferenceService definition
- [x] Add Unit tests to validate arg-based overrides and basic functionality for KServe OIP configuration 
- [x] KServe OIP configuration documentation is updated to include args usage and precedence rules
    - Config precedence is noted: args > environment variables > krausening properties

## Test Strategy/Script
**Set up**
1. Pull down changes and build the project - `mvnd clean install`
2. In Rancher Desktop ensure Traefik is **disabled** - Preferences -> Kubernetes -> Options
3. Download the kserve-example.zip file attached to the ticket and deploy KServe dependencies:
    1. Change to the `kserve-example/deploy` directory and run:
      ```
      helm dependency build ./infrastructure/cert-manager/
      helm dependency build ./infrastructure/kserve-crds/
      helm dependency build ./infrastructure/kserve/
      helm dependency build ./infrastructure/nginx/
      ``` 
    2. Install the services:
      ```
      helm install kserve-crds ./infrastructure/kserve-crds/
      helm install ingress-nginx ./infrastructure/nginx/ --namespace ingress-nginx --create-namespace
      helm install cert-manager ./infrastructure/cert-manager/ --create-namespace --namespace cert-manager
      helm install kserve ./infrastructure/kserve/
      ```
3. Create the namespace: `kubectl create namespace kserve-test`

**Test http**

[...]

PR #121, opened 2025-09-18
#100 Support KServe Args - Kserve Config supports container args

colinpalmer-pro:
- Added arg parsing for KServe Configs so we are in alignment with KServe docs (while also supporting krausening and env variables)
- Added config testing suite

### DIFF
--- a/aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
+++ b/aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
@@ -61,5 +61,4 @@ class AissembleOIPKServe(Model, AissembleOIPService):
             enable_docs_url=self.config.kserve_enable_docs_url,
             enable_latency_logging=self.config.kserve_enable_latency_logging,
             access_log_format=self.config.kserve_access_log_format,
-            grace_period=self.config.kserve_grace_period,
         ).start([self])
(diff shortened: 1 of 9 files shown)


aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/aissemble_oip_kserve.py
```diff
@@ -61,5 +61,4 @@ def start_server(self):
             enable_docs_url=self.config.kserve_enable_docs_url,
             enable_latency_logging=self.config.kserve_enable_latency_logging,
             access_log_format=self.config.kserve_access_log_format,
```
colinpalmer-pro, line 63: i: KServe doesn't expose this for arg overrides so removing

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/config/kserve_args_parser.py
```diff
@@ -0,0 +1,98 @@
...
+# aiSSEMBLE::Open Inference Protocol::Shared
+# %%
+# Copyright (C) 2024 Booz Allen Hamilton Inc.
+# %%
+# This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+# #L%
+###
+import argparse
+
+
```
colinpalmer-pro, line 12: i: Copied `strtobool` from KServe utils to stay aligned to their string to bool parsing

aissemble-open-inference-protocol-shared/tests/features/steps/config_steps.py
```diff
@@ -0,0 +1,130 @@
...
+    pass
+
+
+@given("the configs are set via krausening properties")
+def step_krausening_configs(context):
+    # The properties file location
+    os.environ["KRAUSENING_BASE"] = "tests/resources/krausening/base"
+
+    pm = PropertyManager.get_instance()
+    pm.get_properties("oip.properties", force_reload=True)
```
colinpalmer-pro, line 23: i: Explicitly reloading the properties since PropertyManager didn't use the properties file on the initial test for default values

aaron-gary (approved):

PR #122, opened 2025-09-18
[#110] adding numpy codec

### LINKED ISSUE #110: Feature: Add numpy codec
## Description
Now that #97 is completed to validate that creating a custom codec by following the given documentation in the project is possible we want to incorporate the custom numpy array codec into the project to allow for future users to easily adopt it for their downstream projects.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Add numpy array codec from #97 example downstream project to solution
- [x] Add tests for codec
- [x] Update relevant documentation where necessary

```gherkin
Feature: NumpyCodec encoding and decoding

  Scenario: Decode InferenceRequest with request level content_type
    Given an InferenceRequest with numpy content_type at the request level and payload [1, 2, 3, 4]
    When the payload is decoded
    Then the result should be a numpy array matching [1, 2, 3, 4]

  Scenario: Decode InferenceRequest with input level content_type
    Given an InferenceRequest with numpy content_type at the input level and payload [5, 6, 7, 8]
    When the payload is decoded
    Then the result should be a numpy array matching [5, 6, 7, 8]

  Scenario: Decode InferenceRequest with input level content_type and multiple inputs
    Given an InferenceRequest with numpy content_type at the input level and payload with multiple inputs
    When the payload is decoded
    Then the result should be a multiple outputs with type numpy array

  Scenario: Encode payload into InferenceResponse
    Given a payload with type numpy [9, 10, 11, 12]
    When I encode the payload at the response level
    Then the result should be an InferenceResponse

  Scenario: Encode payload into ResponseOutput
    Given a payload with type numpy [13, 14, 15 ,16]
    When I encode the payload at the request level
    Then the result should be a ResponseOutput
```

## Test Strategy/Script
1. Pull latest version of [aissemble-open-inference-protocol](https://github.com/boozallen/aissemble-open-inference-protocol)
2. Run `mvn clean install -Dmaven.build.cache.skipCache=true -pl :aissemble-open-inference-protocol-shared,aissemble-open-inference-protocol-fastapi`
3. Verify that the build completes successfully
4. Verify documentation references new codec [here](https://github.com/boozallen/aissemble-open-inference-protocol/blob/dev/aissemble-open-inference-protocol-examples/aissemble-oip-shared/README.md)

## Test with sample project
1. Download  and unzip the [110-numpy-codec.zip](https://github.com/user-attachments/files/22431259/110-numpy-codec.zip) folder
2. cd into the unzipped `110-numpy-codec`
3. run `mvn clean install` to install dependencies and set up virtual env
4. add aissemble oip fastapi with the following (be sure to replace `/path/to/` with the actual path to your local version of the project you just built)
   - ```
      poetry add /path/to/aissemble-open-inference-protocol/aissemble-open-inference-protocol-fastapi/dist/aissemble_open_inference_protocol_fastapi-1.1.0.dev0-py3-none-any.whl
     ```
5. start fastapi server by running the following command from the root of the project: `poetry run fastapi dev src/aissemble_oip_fastapi_inference/main.py`
6. View the API docs - http://127.0.0.1:8000/docs
7. Use the Try it out to send the following requests the the /v2/models/{model_name}/infer endpoint
8. Test with `content_type` set at the request-level
**Input:**
```
{
  "id": "test-req",
  "parameters": { "content_type": "numpy" },
  "inputs": [
    {
      "name": "foo",
      "datatype": "INT64",
      "shape": [1, 4],
      "data": [1, 2, 3, 4]
    }
  ]
}
```

**Expected response:**
```
{
  "model_name": "test_model",
  "model_version": null,
  "id": null,
  "parameters": {
    "content_type": "numpy"
  },
  "outputs": [
    {
      "name": "output-0",
      "shape": [
        1,
        4
      ],
      "datatype": "INT64",
      "parameters": {
        "content_type": "numpy"
      },
      "data": [
        1,
        2,
        3,
        4
      ]
    }
  ]
}
```

**Expected log message:**
```
Handler: Request-level codec found and applied -> decoded payload: array([1, 2, 3, 4]) (ndarray of element‐types {'int64'})
```
9. Test with input-level `content_type` only

**Input:**
```
{
  "id": "test-req",
  "inputs": [
    {
      "name": "foo",
      "datatype": "INT64",
      "shape": [1, 4],
      "parameters": { "content_type": "numpy" },
      "data": [1, 2, 3, 4]
    }
  ]
}
```

**Expected response:**
```
{
  "model_name": "test_model",
  "model_version": null,
  "id": null,
  "parameters": null,
  "outputs": [
    {
      "name": "output-0",
      "shape": [
        1,
        4
      ],
      "datatype": "INT64",
      "parameters": {
        "content_type": "numpy"
      },
      "data": [
        1,
        2,
        3,
        4
      ]
    }
  ]
}
```

**Expected log message:**
```
INFO Handler: Falling back to raw InferenceRequest (no request-level codec found)
INFO Handler: Input 'foo' was decoded via per-input codec -> array([1, 2, 3, 4]) (content_type: numpy)
INFO Handler: Returning fallback echo: [array([1, 2, 3, 4])]
```

10. set both request-level and input-level `content_type`

**Input:**
```
{
  "id": "test-req",
  "parameters": { "content_type": "numpy" },
  "inputs": [
    {
      "name": "foo",
      "datatype": "INT64",
      "shape": [1, 4],
      "parameters": { "content_type": "numpy" },
      "data": [1, 2, 3, 4]
    }
  ]
}
```

**Expected response:**
```
{
  "model_name": "test_model",
  "model_version": null,
  "id": null,
  "parameters": {
    "content_type": "numpy"
  },
  "outputs": [
    {
      "name": "output-0",
      "shape": [
        1,
        4
      ],
      "datatype": "INT64",
      "parameters": {
        "content_type": "numpy"
      },
      "data": [
        1,
        2,
        3,
        4
      ]
    }
  ]
}
```

**Expected log message:**
```
INFO Handler: Request-level codec found and applied -> decoded payload: array([1, 2, 3, 4]) (ndarray of element‐types {'int64'})
```

11. Test with multiple inputs (we currently only support a single tensor with `SingleTensorRequestCodec` so we expect to see an error here)
**Input:**
```
{
  "id": "test-req",
  "parameters": { "content_type": "numpy" },
  "inputs": [
    {
      "name": "foo-1",
      "datatype": "INT64",
      "shape": [1, 2],
      "data": [1, 2]
    },
    {
      "name": "foo-2",
      "datatype": "INT64",
      "shape": [1, 2],
      "data": [3, 4]
    }
  ]
}
```

**Expected response:**
```
{
  "model_name": "test_model",
  "model_version": null,
  "id": null,
  "parameters": {
    "content_type": "numpy"
  },
  "outputs": [
    {
      "name": "output-0",
      "shape": [
        1,
        2
      ],
      "datatype": "INT64",
      "parameters": {
        "content_type": "numpy"
      },
      "data": [
        1,
        2
      ]
    }
  ]
}
```

**Expected log message:**
```
ERROR SingleTensorRequestCodec: 'numpy' only supports single input
INFO Handler: Request-level codec found and applied -> decoded payload: array([1, 2]) (ndarray of element‐types {'int64'})
```
12. Tear down server with `CTRL+C`

lemuelwatts:


aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
```diff
@@ -0,0 +1,144 @@
...
+class NumpyArrayCodec(InputCodec):
+    ContentType = "numpy"
+
+    @classmethod
+    def can_encode(cls, payload: Any) -> bool:
+        """
+        Return True if numpy codec can handle the given payload.
+        """
+        try:
+            arr = to_numpy_array(payload)
```
cwoods-cpointe, line 103: a: seems excessive to convert it to a numpy array if we are only checking if its possible. Probably fine but would be worth seeing if there is something simpler.

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
```diff
@@ -0,0 +1,144 @@
...
+            return arr.ndim >= 1
+        except Exception:
+            return False
+
+    @classmethod
+    def encode_input(cls, name: str, payload: Any, **kwargs) -> RequestInput:
+        """
+        Encode a Python array into a RequestInput.
+        """
+        output = cls.encode_output(name=name, payload=payload)
```
cwoods-cpointe: a: a little confusing that encode_input is calling encode_output. Could make more sense to have a separate encode_payload function that they both call

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
```diff
@@ -0,0 +1,144 @@
...
+            parameters=output.parameters,
+        )
+
+    @classmethod
+    def encode_output(cls, name: str, payload: Any, **kwargs) -> ResponseOutput:
+        """
+        Encode a Python array into a ResponseOutput.
+        """
+        arr = to_numpy_array(payload)
+        shape = [len(payload), 1]
```
cwoods-cpointe: q: having a hard time tracking the flow for multidimensional arrays. Does it flatten them down to one dimension?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
```diff
@@ -0,0 +1,144 @@
...
+def to_numpy_array(arr: Any, dtype: np.dtype = None) -> np.ndarray:
+    # Convert to numpy array if not already one with appropriate dtype
+    if isinstance(arr, np.ndarray):
+        return arr
+    return np.array(arr, dtype=dtype)
+
+
+def decode_input_or_output(input_or_output: InputOrOutput) -> np.ndarray:
+    # Decode the data from RequestInput or ResponseOutput into a numpy array
+    # Also determines the appropriate numpy dtype based on the protocol datatype
```
cwoods-cpointe: a: can put these comments as docstrings like the classmethods are doing

aissemble-open-inference-protocol-shared/tests/features/steps/numpy_codec_steps.py
```diff
@@ -0,0 +1,157 @@
...
+
+@given(
+    "an InferenceRequest with numpy content_type at the request level and payload {payload}"
+)
+def step_given_request_level_numpy(context, payload):
+    data = ast.literal_eval(payload)
+    context.expected = np.array(data)
+    input_obj = RequestInput(
+        name="input-0",
+        datatype="INT64",
```
cwoods-cpointe: a: datatype is not a string object

aissemble-open-inference-protocol-shared/tests/features/steps/numpy_codec_steps.py
```diff
@@ -0,0 +1,157 @@
...
+        shape=[len(data), 1],
+        data=data,
+        parameters=Parameters(),
+    )
+    context.inference_request = InferenceRequest(
+        model_name="test-model",
```
cwoods-cpointe: a: model_name is not an attribute of inferenceRequest

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
```diff
@@ -0,0 +1,144 @@
...
+            data=output.data,
+            parameters=output.parameters,
+        )
+
+    @classmethod
+    def encode_output(cls, name: str, payload: Any, **kwargs) -> ResponseOutput:
+        """
+        Encode a Python array into a ResponseOutput.
+        """
+        arr = to_numpy_array(payload)
```
colinpalmer-pro, line 108: Q/C: On the shape assignment, if we had a multi-dimensional array, I think this would cause an issue.
e.g. 3x3 array would get shape = [3, 1] (claiming 3 rows, 1 column) but the actual data would have 9 elements.

Maybe it should be `shape = list(arr.shape)` instead?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
colinpalmer-pro: Moving my comment here since Clay beat me to it:

Q/C: On the shape assignment, if we had a multi-dimensional array, I think this would cause an issue.
e.g. 3x3 array would get shape = [3, 1] (claiming 3 rows, 1 column) but the actual data would have 9 elements.

Maybe it should be `shape = list(arr.shape)` instead?

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
lemuelwatts: refactored so they both call a shared `encode_payload` function

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
lemuelwatts: ah yeah good catch. modified so we use `shape = list(arr.shape)` instead

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
lemuelwatts: updated to use docstrings on functions

aissemble-open-inference-protocol-shared/tests/features/steps/numpy_codec_steps.py
lemuelwatts: removed all instances of `model_name` on inferenceRequests

aissemble-open-inference-protocol-shared/tests/features/steps/numpy_codec_steps.py
lemuelwatts: modified to use correct Datatype object

aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/codecs/numpy.py
lemuelwatts: Yeah I agree it feels a little strange to do the whole conversion but I couldn't find another clean way to do it without doing a bunch of `isinstance` checks and checking the attributes of the payload

cwoods-cpointe (approved):
Can we add a test scenario for multi dimensional arrays?
Other than that, looks good

lemuelwatts:
> Can we add a test scenario for multi dimensional arrays? Other than that, looks good

Added multi-dimensional array test!

Issue #108, opened 2025-09-11
Feature: Consolidate custom Dataplane implementation for all platform


jaebchoi:
## Description

We currently have two DataplaneHandler for Grpc,FastAPI vs KServe as Kserve need to extend their own version of Dataplane Class. Therefore, it results in having user to implement separate Dataplane handler if they want to switch from FastAPI or gRPC into KServe. In order to reduce this extra steps we can consolidate this work by calling our Shared DataplaneHandler inside of KServeDataplaneHandler implementation.

## DOD
- [x] Consolidate Dataplane by calling Shared DataplaneHandler inside of KServeDataplaneHandler implementation.
    - [x] KServeDataplaneHandler will act as Adapter which wraps [DataplaneHandler](https://github.com/boozallen/aissemble-open-inference-protocol/blob/dev/aissemble-open-inference-protocol-shared/src/aissemble_open_inference_protocol_shared/handlers/dataplane.py) so that KServeDataplaneHandler logic does what DataplaneHandler does
- [x] Add unit test to verify above logic.

## Test Strategy/Script
- Pull latest [feature/kserve-custom-handler](https://github.com/boozallen/aissemble-open-inference-protocol/tree/feature/kserve-custom-handler) branch build the project - `mvnd clean install`
- Download [kserve-example.zip](https://github.com/user-attachments/files/22238326/kserve-example.zip)
- unzip and cd into `kserve-example/deploy` directory and run:
```
helm dependency build ./infrastructure/cert-manager/
helm dependency build ./infrastructure/kserve-crds/
helm dependency build ./infrastructure/kserve/
helm dependency build ./infrastructure/nginx/
``` 
- Install the services:
```
helm install kserve-crds ./infrastructure/kserve-crds/
helm install ingress-nginx ./infrastructure/nginx/ --namespace ingress-nginx --create-namespace
helm install cert-manager ./infrastructure/cert-manager/ --create-namespace --namespace cert-manager
helm install kserve ./infrastructure/kserve/
```
- Create the namespace: `kubectl create namespace kserve-test`

### Test Deploying KServe
- cd into kserve deployment charts 
` cd aissemble-open-inference-protocol-examples/aissemble-oip-kserve/aissemble-oip-kserve-containerization/aissemble-oip-kserve-containerization-deploy/src/main/resources/templates`
- Apply Inference service helm chart to kserve-test namespace
` kubectl apply -n kserve-test -f ./inference-service.yaml`
[...]

PR #125, opened 2025-09-19
Feature/kserve custom handler

jaebchoi:
This PR includes issue [92](https://github.com/boozallen/aissemble-open-inference-protocol/issues/92) and [108](https://github.com/boozallen/aissemble-open-inference-protocol/issues/108)


aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/kserve_dataplane.py
```diff
@@ -0,0 +1,135 @@
...
+from aissemble_open_inference_protocol_kserve.mappers.infer_mapper import InferMapper
+
+from aissemble_open_inference_protocol_shared.handlers.dataplane import (
+    DataplaneHandler,
+    DefaultHandler,
+)
+
+
+class KServeDataplaneHandler(DataPlane):
+    """KServe Custom DataPlane Handler"""
```
epwilkins: A: We should probably call this an Adapter so it's more easily understood in the context of typical design patterns.  I'd also fill out the docs to explain that it converts aissemble-oip's interface to kserve's interface.

epwilkins (approved):

aissemble-open-inference-protocol-kserve/src/aissemble_open_inference_protocol_kserve/kserve_dataplane.py
jaebchoi: sounds good Ill rename to add Adapter and add corresponding update

Issue #132, opened 2025-09-29
TASK: Align aiSSEMBLE OIP Kserve and FastAPI uvicorn versions


cwoods-cpointe:
## Description
Our FastAPI implementation takes in a uvicorn version that is incompatible with the Kserve implementation. If users are moving towards a Kserve solution it is unlikely that they will still need the FastAPI library. Regardless we should align these versions

## DOD
- [x] aiSSEMBLE OIP FastAPI and Kserve are using the same uvicorn verisons
    - [x] FastAPI uvicorn dep is pinned to `>=0.30.6,<0.31.0`

## Test Strategy/Script
1. Navigate to the Fastapi getting started example - `cd aissemble-open-inference-protocol-examples/aissemble-oip-fastapi/aissemble-oip-fastapi-inference`
2. Build - `mvn clean install`
3. Start the app - `poetry run run_server`
4. Send a server ready request
```
curl -w "\nHTTP Code: %{http_code}\n" \
-H 'Content-Type: application/json' \
-X GET \
http://127.0.0.1:8082/v2/health/ready
```
6. Verify it returns successfully and the server is up

cwoods-cpointe:
DoD and OTS completed with @csun-cpointe

cwoods-cpointe:
Decided to mark this effort as OBE for now. There isnt a real use case for users to have both Kserve and Fastapi/gRPC solutions running in the same project because Kserve handles the server deployments.

PR #134, opened 2025-09-30
[#132] Pin uvicorn version

cwoods-cpointe:
Set the uvicorn version to the same version that Kserve is using. Helps users transition from our Fastapi solution to Kserve

### DIFF
--- a/aissemble-open-inference-protocol-fastapi/pyproject.toml
+++ b/aissemble-open-inference-protocol-fastapi/pyproject.toml
@@ -34,7 +34,8 @@ pydantic = "^2.11.4"
 PyJWT="^2.10.1"
 krausening = ">=20"
 cryptography = ">=44.0.1"
-uvicorn = ">=0.35.0"
+# Pinned to this version to stay compatible with the aiSSEMBLE OIP Kserve implementation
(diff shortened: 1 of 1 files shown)


aissemble-open-inference-protocol-fastapi/pyproject.toml
```diff
@@ -34,7 +34,8 @@ pydantic = "^2.11.4"
 PyJWT="^2.10.1"
 krausening = ">=20"
 cryptography = ">=44.0.1"
-uvicorn = ">=0.35.0"
+# Pinned to this version to stay compatible with the aiSSEMBLE OIP Kserve implementation
+uvicorn = ">=0.30.6,<0.31.0"
```
epwilkins, line 38: A: We should consider using Habushu's [managedDependencies](https://github.com/TechnologyBrewery/habushu/blob/dev/docs/CONFIGURATION_README.md#managed-dependencies-configurations)

epwilkins (approved):

cwoods-cpointe:
Decided to mark this effort as OBE for now. There isnt a real use case for users to have both Kserve and Fastapi/gRPC solutions running in the same project because Kserve handles the server deployments.

PR #144, opened 2025-12-18
143 initial client layout

d-ryan-ashcraft:


aissemble-oip-core/src/aissemble_oip/client/inference_client.py
```diff
@@ -0,0 +1,73 @@
...
+    #
+    # def summarize(self, model_name: str) -> SummarizationBuilder:
+    #     """
+    #     Creates a builder for summarization inference.
+    #
+    #     :param model_name: The name of the model to use.
+    #     :return: A SummarizationBuilder instance.
+    #     """
+    #     # TODO: Implement summarization builder creation
+    #     pass
```
lemuelwatts: A: do we want to leave these commented out for now?

lemuelwatts (approved):

aissemble-oip-core/src/aissemble_oip/client/inference_client.py
d-ryan-ashcraft: I commented them out because `ObjectDetectionBuilder` and `SummarizationBuilder` don't yet exist

lemuelwatts: ah makes sense to me!

aissemble-oip-core/src/aissemble_oip/client/builder/inference_builder.py
```diff
@@ -0,0 +1,206 @@
...
+    def streaming(self) -> bool:
+        """Whether the inference should be performed in streaming mode."""
+        return self._streaming
+
+    # -------------------------------------------------------------------------
+    # Fluent configuration methods (to be implemented/chained by subclasses)
+    # -------------------------------------------------------------------------
+
+    def with_model(self, model_name: str) -> "InferenceBuilder":
+        """
```
ekoniec1: A/Q: Do we want to adopt NumPy style docstrings for this codebase?  IDEs should be able to handle whatever and consistency is of course the most important, but I'd make the argument that since Google style docstrings are more terse, they're a better fit for this type of project.

aissemble-oip-core/src/aissemble_oip/client/builder/raw_inference_builder.py
ekoniec1, line 1: A: Typo in `raw_inference_buidler.py`

aissemble-oip-core/src/aissemble_oip/client/builder/inference_builder.py
```diff
@@ -0,0 +1,206 @@
...
+            Self, for method chaining.
+        """
+        self._streaming = True
+        return self
+
+    # -------------------------------------------------------------------------
+    # Iterator protocol support (for streaming responses)
+    # -------------------------------------------------------------------------
+
+    def __iter__(self) -> Generator[Any, None, None]:
```
ekoniec1, line 164: Q: I would have thought that the generator would be somewhere else, like on the `Predictor`? This implies that the `InferenceBuilder` is yielding inference results?

ekoniec1 (approved):

aissemble-oip-core/src/aissemble_oip/client/builder/inference_builder.py
d-ryan-ashcraft: Yeah - I think that's fair.  I was playing with some different coding agents and that's probably where the diff came from

aissemble-oip-core/src/aissemble_oip/client/builder/inference_builder.py
d-ryan-ashcraft: I was trying to say that this is a python generator for iterating - but not sure I did a great job of specifying that.

We may also ultimately move or change it, but here it's more about iterating over results rather than a stream from the server side.

aissemble-oip-core/src/aissemble_oip/client/builder/raw_inference_builder.py
d-ryan-ashcraft: ✅ Good catch - fixed!

aissemble-oip-core/src/aissemble_oip/client/builder/inference_builder.py
d-ryan-ashcraft: ✅ Resolved

Issue #143, opened 2025-12-17
Feature: Add foundational components for OIP client


d-ryan-ashcraft:
## Description
Start the implementation of the client with foundational classes and stubbed out implementations so that we can start to devise and concur on the work more easily.

The client design is captured in #141

PR #145, opened 2025-12-18
[#143] :green_heart: fixed underspecified build configuration and dow…

d-ryan-ashcraft:
…nstream items results from the underspecification


aissemble-oip-core/pom.xml
```diff
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>aissemble-oip-core</artifactId>
-    <packaging>pom</packaging>
```
d-ryan-ashcraft, line 12: I (Informational): I had accidentally rolled back this change earlier 🤦... but at least caught it during my regression test of the build.  The subsequent changes in this PR are downstream from getting the build working again.

aissemble-oip-core/pyproject.toml
```diff
@@ -9,6 +9,12 @@ authors = [
 requires-python = ">=3.9.14"
 dependencies = []
 
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.ruff.lint]
+ignore = ["E741"]
```
d-ryan-ashcraft, line 17: This rule doesn't like the generic type variables `I` and `O` because it thinks they will be confused for `0` and `1` (the numbers).  

When we edited in text editors, this was a reasonable concern.  Hard to stomach and make them not match normal generic type naming conventions for such an outdated PEP 8 standard, so I'm turning it off.

taycavett98 (approved):

Issue #150, opened 2026-01-14
Feature: create sample problem for object detection


d-ryan-ashcraft:
## Description
Establish an example project that run's the object detection code base as a real example.

## DOD
- [x] Create `aissemble-oip-examples/aissemble-object-detection-example`
- [x] Ensure it runs with MLServer, preferably as a unit test
- [x] Refine less optimal or missing parts of the implementation via this process

PR #151, opened 2026-01-15
[#150] :sparkles: added yolo example; :building_construction: refacto…

d-ryan-ashcraft:
…ring to support a yolo module, inline with the original design


aissemble-oip-examples/aissemble-object-detection-example/tests/features/environment.py
d-ryan-ashcraft, line 1: I (Informational): probably worth splitting this out to a separate common test module, but will defer this until we add more examples.

lemuelwatts (approved):

Issue #152, opened 2026-01-15
Feature: Add summarization option to core


d-ryan-ashcraft:
## Description
Support the summarization API in `aissemble-oip-core`

## Definition of Done
- [x] Add module
- [x] Add simple BDD test
- [x] Ensure build works

## Test Steps
- [x] Run the following and ensure it completes successfully: `mvn clean install`

d-ryan-ashcraft:
CI build passed:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for aiSSEMBLE::Open Inference Protocol 1.5.0-SNAPSHOT:
[INFO] 
[INFO] aiSSEMBLE::Open Inference Protocol ................. SUCCESS [  9.445 s]
[INFO] aiSSEMBLE::Open Inference Protocol::Core ........... SUCCESS [ 12.656 s]
[INFO] aiSSEMBLE::Open Inference Protocol::Modules ........ SUCCESS [ 13.465 s]
[INFO] aiSSEMBLE::Open Inference Protocol::Modules::YOLO .. SUCCESS [01:14 min]
[INFO] aiSSEMBLE::Open Inference Protocol::Examples ....... SUCCESS [  7.141 s]
[INFO] aiSSEMBLE::Open Inference Protocol::Examples::Object Detection SUCCESS [ 21.275 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:18 min
[INFO] Finished at: 2026-01-16T14:25:31Z
[INFO] ------------------------------------------------------------------------
```

PR #153, opened 2026-01-16
[#152] :sparkles: initial summarization implementation within aisseml…

d-ryan-ashcraft:
…e-oip-core only


aissemble-oip-modules/aissemble-oip-yolo/src/aissemble_oip_yolo/runtime.py
```diff
@@ -71,6 +71,17 @@ async def load(self) -> bool:
         self.ready = True
         return self.ready
```
d-ryan-ashcraft, line 73: I (Informational): updates to support some inconsistent results with regression test runs

taycavett98 (approved):

Issue #164, opened 2026-01-23
Feature: add docker deployment option


d-ryan-ashcraft:
## Description
Continue rounding out deploy functionality with a docker option

## DOD
- [x] Code docker option
- [x] Ensure it reuses dependencies from `uv` build
- [x] Update documentation

## Test Strategy/Script
### Prerequisites                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  - Docker Desktop or Rancher Desktop running                                                                                                                                                                                            
  - uv installed (uv --version)                                                                                                                                                                                                          
  - Project built (mvn clean install from root)                                                                                                                                                                                          
                                                                                                                                                                                                                                         
### Test Steps                                                                                                                                                                                                                                                                                                                                                                                                                                                            
1. Navigate to the summarization example and install needed dependencies:
```                                                                                                                                                                                   
cd aissemble-oip-examples/aissemble-summarization-example
pip install ../../aissemble-oip-deploy/dist/aissemble_oip_deploy-1.5.0.dev0-py3-none-any.whl
```

2. Generate Docker deployment configs:
```
oip deploy init --target docker
```                                                                                                                                                                                                                                    
- Expected output:                                                                                                                                                                                                                       
  - Message showing models found (sumy-textrank, sumy-lsa)                                                                                                                                                                               
  - Message about updating pyproject.toml                                                                                                                                                                                                
  - Message about running uv lock                                                                                                                                                                                                        
  - Generated files listed (Dockerfile, docker-compose.yml, .dockerignore, README.md)                                                                                                                                                    


3. Build and run the container:
```
  cd deploy/docker
  docker-compose up --build
```                                                                                                                                                                                                                                         
-Expected: Container builds successfully and MLServer starts, showing models loaded

4. Test the endpoint:
- Open http://127.0.0.1:8080/v2/docs
- Find the `/v2/models/{model_name}/infer` endpoint
- Populate the following fields:values:
  -`model_name`: `sumy-textrank`
  - `request body`:
```
{
  "inputs": [
    {
      "name": "text",
      "shape": [
        1
      ],
      "datatype": "BYTES",
      "data": [
[...]

PR #166, opened 2026-01-26
[#164] :sparkles: prior commit missed this test and PR feedback - thi…

d-ryan-ashcraft:
…s one really enable docker support


aissemble-oip-deploy/src/aissemble_oip_deploy/templates/docker/Dockerfile.j2
```diff
@@ -1,20 +1,35 @@
...
-# Copy dependency files
-COPY pyproject.toml uv.lock ./
+{% if use_wheels %}
+# Copy pre-built wheels with pinned dependencies (dev mode)
+COPY deploy/docker/wheels/ /wheels/
 
-# Install runtime dependencies into .venv
-RUN uv sync --frozen --no-dev --group runtime
+# Create virtual environment, install mlserver from PyPI, then local wheels
+RUN uv venv .venv && \
```
d-ryan-ashcraft, line 22: S: mlserver is installed from PyPI first, then local wheels are installed. If local wheels have `mlserver>=X.Y.Z` as a dependency with version constraints, this could cause version conflicts or downgrades. The wheels built by `uv-monorepo-dependency-tool` have pinned dependencies, and the order of installation matters.                                                                                                                                           
                                                                                                                                                                                                                                         
Suggestion: Install wheels first (which will pull in `mlserver` as a dependency), or better yet, include `mlserver` in the requirements derived from the wheels.

aissemble-oip-deploy/src/aissemble_oip_deploy/cli.py
```diff
@@ -135,6 +135,33 @@ def init(
     project_dir = project_dir or Path.cwd()
     output_dir = output_dir or project_dir / "deploy"
 
+    # Check if running from wrong directory (inside deploy/)
```
d-ryan-ashcraft, line 138: S (Substantive): This check will incorrectly trigger for any path containing a directory named "deploy" anywhere, e.g., `/projects/deploy-toolkit/models/`or `/home/user/deploy-manager/`. The intent is to catch users who cd'd into their project's `deploy/` output directory, but the check is too broad.                                                                                                                                                                        
                                                                                                                                                                                                                                         
Suggestion: Check specifically for the pattern where the current directory is named deploy or ends with `/deploy`, not just any path containing "deploy" as a component.

aissemble-oip-deploy/src/aissemble_oip_deploy/templates/docker/Dockerfile.j2
d-ryan-ashcraft: ✅ Removed the separate `mlserver` install. The wheels will bring in `mlserver` as a transitive dependency. This also lets `uv`'s resolver handle all dependencies together, avoiding potential version conflicts.

aissemble-oip-deploy/src/aissemble_oip_deploy/cli.py
d-ryan-ashcraft: Changed "deploy" in cwd.parts to cwd.name == "deploy". This now only triggers when the current directory is literally named "deploy", not when "deploy" appears anywhere in the path like `/projects/deploy-toolkit/`.

This is probably good enough, but we'll let time tell.

